Issue #8, opened 2022-10-25
 Setup language switch and fix bug


Albertlarsen:
We need a way to set chosen language for a user and a place to get localizations based on key.

Also fix bug: Getting error message when trying to use localized words in frontend.

PR #17, opened 2022-11-07
Feat/setup language switch

Albertlarsen:
## Description
Fixes ts errors when using translation service and makes app able to switch language.

## Related Issue(s)
- #8 

## Verification
- [ ] All tests run green

## Documentation


src/main.jsx
```diff
@@ -31,10 +32,21 @@ const router = createBrowserRouter([
   },
 ]);
 
+const initLanguage = (lang) => {
+  if (lang === 'no') {
```
allinox, line 36: Should probably use 'nb' instead as 'no' referst to norwegian as a whole. When there is the distiction between nynorsk and bokmål, I think we should use 'nn' and 'nb' respectively

allinox (approved):
Good stuff 😊

Issue #23, opened 2022-11-23
Create API-Delegation view S4 | NewOrgDelegationPage


Albertlarsen:
### Description

As a tjenesteeier I want to to choose an API from a list so I can delegate API-access to that org.

### Acceptance Criterias

AC:

- [ ] "Add"-functionality works
- [ ] "Remove"-functionality works
- [ ] Multiselect and filter works
- [ ] Chosen API-list works

PR #31, opened 2022-12-01
Feat/api delegation s4 new org delegation page

Albertlarsen:
## Description
An implementation of the most components for NewOrgDelegationPage while waiting for UX design for "add new org to list"-functionality. There's also been created a new reusable Accordion: NewDelegationAccordion.

## Related Issue(s)
- #23 

## Verification
- [ ] All tests run green

## Documentation


src/components/Common/NewDelegationAccordion/NewDelegationAccordion.tsx
```diff
@@ -0,0 +1,91 @@
...
+  Remove = 'remove',
+}
+
+export interface NewDelegationAccordionProps {
+  title: string;
+  subtitle: string;
+  hasOrgNr?: boolean;
+  description: string;
+  buttonType: NewDelegationAccordionButtonType;
+  callback: () => { payload: DelegableOrg | DelegableApi; type: string };
```
allinox, line 30: Another thing here: Does the component need to know what the payload of the callback is here? The parameters aren't used here, they're baked into the callback outside of the component.

So can't this callback-parameter be more openly defined here anyway?

src/components/NewOrgDelegationPage/NewOrgDelegationPage.module.css
```diff
@@ -0,0 +1,59 @@
+@media only screen and (min-width: 0px) {
+  .page{
+      --pageContainer-width: 1055px;
+  }
+}
+
+@media only screen and (min-width: 700px) {
+  .page{
+    --pageContainer-width: 1055px;
```
allinox, line 9: Why specify for smaller screens if the widths are all the same anyway?

src/components/Common/NewDelegationAccordion/NewDelegationAccordion.tsx
Albertlarsen: If we use any we will receive lint error and not using anything gives typescript error. I also think defining expected payload-object gives value. I don't know what you mean by callback-parameter being more openly defined.

src/components/NewOrgDelegationPage/NewOrgDelegationPage.module.css
Albertlarsen: Because we have to write these exact media queries in a few weeks :P

src/components/Common/NewDelegationAccordion/NewDelegationAccordion.tsx
allinox: I mean that you can do like [here](https://github.com/Altinn/altinn-access-management-frontend/blob/025ecc80b05e785b8017c698b2895a62d9573023/src/components/ApiDelegationOverviewPage/OrgDelegationOverviewPageContent/OrgDelegationAccordion/DeletableListItem/DeletableListItem.tsx#L12)

NewDelegationAccordion does not inherently need to be related to Orgs or Apis, and it doesn't need to know whether or not your callback is related to DelegableOrg or DelegableApi or what you do with them. It just needs to call the function you give it.

allinox: At least that's how I see it.
I don't know what value these payload-objects would give here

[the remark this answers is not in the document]
allinox: Or onActionButton? I mean, technically, clicking the accordion (as in opening it) is an action too, and that's not what we're talking about here. But the header isn't an ActionButton.
Or maybe it is? Not sure tbh...

Albertlarsen: You're right, I forgot it could be done by returning void.

src/components/NewOrgDelegationPage/NewOrgDelegationPage.module.css
allinox: Ahaaaa... So they're just placeholders then?

Albertlarsen: Yup, you can choose if we should keep or remove them.

allinox: They might as well be there 🙂

src/components/Common/NewDelegationAccordion/NewDelegationAccordion.tsx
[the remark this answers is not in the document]
Albertlarsen: Yeah maybe addRemoveClick to avoid ambiguity then

allinox (approved):
Good stuff 😊

PR #52, opened 2022-12-14
Feat/create s5a api delegation confirmation page

Albertlarsen:
## Description
Create resuable components ConfirmationPage and compactDeletableListItem and make api delegation confirmation page use it.

## Related Issue(s)
#65 

## Verification
- [ ] All tests run green

## Documentation


src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+  removeCallback?: (() => void) | null;
+  firstText: string;
+  secondText: string;
+  startIcon?: React.ReactNode;
+}
+
+export const CompactDeletableListItem = ({
+  removeCallback,
+  firstText,
+  secondText,
```
allinox, line 24: Use left- and rightText insted? Or primary and secondary text/label?

src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+  startIcon,
```
allinox, line 25: Wouln't Icon be sufficient?

src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+  ButtonColor,
+  SvgIcon,
+} from '@altinn/altinn-design-system';
+import { useTranslation } from 'react-i18next';
+
+import { ReactComponent as MinusCircle } from '@/assets/MinusCircle.svg';
+
+import classes from './CompactDeletableListItem.module.css';
+
+export interface DeletableListItemProps {
```
allinox, line 14: Could it be problematic that these props have the same name as the other (non-compact) DeletableListItem when they don't contain the same props?
Maybe rename this one to CompactDeletableListItemProps?

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { CompactDeletableListItem } from '@/components/Reusables/CompactDeletableListItem';
+
+import classes from './ConfirmationPage.module.css';
+
+export interface ConfirmationPageProps {
+  firstListItems: DelegableApi[];
+  secondListItems: DelegableOrg[];
+  headerText: string;
+  firstContentText?: string;
+  secondContentText?: string;
```
allinox, line 27: I don't know if first- and secondContentText are optimal names here, not when they're both actually headers for the first and second list respectively.

Actually, since the first list is explicitly coded to contain Api's and the second is coded to contain Orgs, maybe the props for their headers could be named apiListHeader and orgListHeader?

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+  bottomText?: string;
+  mainButton?: React.ReactNode;
+  complementaryButton?: React.ReactNode;
```
allinox, line 30: Why is this called complimentaryButton when it is of class previousButton?

Can't it just be a backButton or previousButton?

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+import type { DelegableOrg } from '@/rtk/features/delegableOrg/delegableOrgSlice';
+import { ReactComponent as OfficeIcon } from '@/assets/Office1.svg';
+import { ReactComponent as SettingsIcon } from '@/assets/Settings.svg';
+import { CompactDeletableListItem } from '@/components/Reusables/CompactDeletableListItem';
+
+import classes from './ConfirmationPage.module.css';
+
+export interface ConfirmationPageProps {
+  firstListItems: DelegableApi[];
+  secondListItems: DelegableOrg[];
```
allinox, line 24: Since these are of type DelegatableApi and DelegatableOrg, can't they just be called apiList and orgList (I don't actually think you need the Items-part)

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+  headerText: string;
```
allinox, line 25: Maybe call this pageHeader?

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <Page color={color}>
+          <PageHeader icon={headerIcon}>{headerText}</PageHeader>
+          <PageContent>
+            <div className={classes.pageContent}>
+              <h2>{firstContentText}</h2>
+              <List borderStyle={BorderStyle.Dashed}>{apiListItems}</List>
+              <h2 className={classes.secondText}>{secondContentText}</h2>
+              <List borderStyle={BorderStyle.Dashed}>{orgListItems}</List>
+              <h3 className={classes.bottomText}>{bottomText}</h3>
+              <div className={classes.navButtonContainer}>
```
allinox, line 85: Why are the buttons placed in the middle of the page?
image

src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
Albertlarsen: I think the prop should say something about where it the icon is placed

src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
Albertlarsen: Yes, good catch

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
Albertlarsen: Thought we could generalize it if we want it to be something else than backButton in the future

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
Albertlarsen: Hanne wanted them there

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
allinox: But in that case, the code should probably be generalized too.
Right now it has a classname of previousButton and it lies in a dev called navButtonSection.

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
allinox: But the design isn't like that though. The design has them on the left (which I also disagree with for that matter, since I think they should be on the right, but that's another question entirely)

allinox: But if Hanne has approved this, then I suppose it's right

allinox: Oh, but one things though:
Maybe add some more space under the buttons? (And maybe move them up a smidge?)

Right now it looks a bit cramped and doesn't follow framing rules.
(As I look at the design sketches I notice there's more space there too)
image

src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
allinox: Does it need to when there is only one icon?
The Page-component doesn't specify this (and I wouldn't expect it to either)

src/components/Reusables/ConfirmationPage/ConfirmationPage.tsx
allinox: Aaaahh, that makes sense. It's all good then 😊

src/components/Reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
Albertlarsen: Maybe not, I'll change it to icon

Issue #58, opened 2022-12-19
Make x-button and fix routing on it


Albertlarsen:
Make button in pic below in a good way so it's applied to all API-delegering Pages in a good way from 1 place. Route: https://www.altinn.no/ui/Profile.
<img width="776" alt="image" src="https://user-images.githubusercontent.com/31471142/208451706-807dea38-a483-42e3-9eb8-3880b6af8f9d.png">

AC:
- The button takes you to https://www.altinn.no/ui/Profile.

PR #79, opened 2022-12-30
Feat/make exit button

Albertlarsen:
## Description
In this PR there's implemented a new component called PageContainer that sets correct style for Pages and adds an exit-button. This is to sentralize stuff like this and make sure you only need to make a change one place for it to be applied on all Pages. The usage of the component is also implemented on all current pages.

## Related Issue(s)
- #58 

## Verification
- [ ] All tests run green

## Documentation


src/components/Reusables/PageContainer/PageContainer.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <div className={classes.pageWrapper}>
+      <div className={classes.exitButton}>
+        <Button
+          variant={ButtonVariant.Quiet}
+          color={ButtonColor.Inverted}
+          size={ButtonSize.Medium}
+          icon={<ExitIcon />}
+          onClick={() => navigate('/Profile')}
+        ></Button>
```
allinox, line 24: Is this the right icon?
image
image

src/components/Reusables/PageContainer/PageContainer.module.css
```diff
@@ -0,0 +1,12 @@
+.pageWrapper{
+  width: 1055px;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 1rem;;
+  background-color: #1EADF7;
```
allinox, line 6: Should this be a token from the design system or something?

Albertlarsen: The figma tokens are only supposed to be consumed by the design system for now. Maybe in future we can base our color on the figma tokens when they are ready to be consumed by other altinn apps.

Issue #117, opened 2023-01-31
Update axios calls to use cookie values


Albertlarsen:
Currently axios calls doesn't use cookie values. This task is about updating all those calls to make them use cookie values.

Tasks: 

- [x] Find a way to get cookies.
- [x] update axios calls to use cookies in order to do axios calls.

AC:

- Check if returned data corresponds to the data that's expected to be returned.

sivaglen:
@Albertlarsen trenger vi dette for å ferdigstille delegerbare API ressurser?

Albertlarsen:
@sivaglen Ja men det tar ikke lang tid

PR #120, opened 2023-02-01
Make axios calls use cookie

Albertlarsen:
## Description
This PR makes axios calls use altinnPartyId from cookie.

## Related Issue(s)
- #117 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/resources/Cookie/CookieMethods.ts
+++ b/src/resources/Cookie/CookieMethods.ts
@@ -1,0 +1,15 @@
+export function getCookie(cname: string) {
+  const name = cname + '=';
+  const decodedCookie = decodeURIComponent(document.cookie);
+  const ca = decodedCookie.split(';');
+  for (let i = 0; i < ca.length; i++) {
+    let c = ca[i];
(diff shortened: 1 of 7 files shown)


src/resources/Cookie/CookieMethods.ts
```diff
@@ -0,0 +1,15 @@
+export function getCookie(cname: string) {
+  const name = cname + '=';
+  const decodedCookie = decodeURIComponent(document.cookie);
+  const ca = decodedCookie.split(';');
+  for (let i = 0; i < ca.length; i++) {
+    let c = ca[i];
```
allinox, line 6: Couldn't this while underneath be replaced by a trim()?
```suggestion
    let c = ca[i].trim();
```

allinox: Maybe.. Not sure how much time would be saved in this case, and there is an argument to be made about readability, but this _is_ only nitpicking so it doesn't really matter that much.
It's fine as it is 🙂

allinox (approved):

Issue #63, opened 2022-12-16
API-delegering | S5a/S5b | Integrate with backend to actually delegate apis to org


Albertlarsen:
This task is about delegating api-accesses when clicking on the "bekreft delegering"-button.

Additional info:

- I've created an inspiration on how to call the delegation endpoint in method **delegateApi** in **delegationRequestSlice**. That method was created to take care of we only could do a delegation one by one in the endpoint, not batchDelegations which I suspect Jon Kjetil has made. 
- Talk with Jon Kjetil about the endpoint he's made.

Tasks:

- [x] Hook "bekreft delegering"-button up with backend call. 
- [x] Make sure delegation happens
- [x] Make sure AS-XSRF-TOKEN is pasted in the axios call in "bekreft delegering".
- [x] Update receipt component so it reflects the successes and failures of the delegations
- [x] Reload oveview after delegation to display new entry
- [x] Use org number to delegate to chosen org 

AC:

- "bekreft delegering"-button sends you to receipt page and succesful and failed delegations is returned.

![Image](https://user-images.githubusercontent.com/31471142/211568113-99e23c02-6997-42da-8805-e7028d4bd7a0.png)

![Image](https://user-images.githubusercontent.com/31471142/216579585-b391d553-65d0-4195-9538-3c0d1b0f8b17.png)

PR #149, opened 2023-02-15
Delegate apis through backend

allinox:
## Description

Loop through chosen arpi's and orgs and make delegations using backend call.
Fixes to summary page component to display successful and/or failed delegations.
Local running requires setup for backend.

## Related Issue(s)
- #63

## Verification
- [ ] All tests run green

## Documentation


src/rtk/features/overviewOrg/overviewOrgSlice.ts
```diff
@@ -148,14 +148,13 @@ const createCopyOrg = (org: OverviewOrg) => {
   };
 };
 
-export const fetchOverviewOrgsOutbound = createAsyncThunk(
-  'overviewOrg/fetchOverviewOrgsOutbound',
+export const fetchOverviewOrgsOffered = createAsyncThunk(
```
Albertlarsen, line 151: Either use same vocabulary as introduced in apiDelegation/given or change the folder name from apiDelegation/given => apiDelegation/offered

```suggestion
export const fetchOverviewOrgsGiven = createAsyncThunk(
```

src/rtk/features/delegationRequest/delegationRequestSlice.ts
```diff
@@ -1,86 +1,111 @@
...
+export const postApiDelegation = createAsyncThunk(
+  'delegationRequestSlice/postApiDelegation',
+  async (delegationInfo: DelegationRequest) => {
+    const { apiIdentifier, apiName, orgNr, orgName }: DelegationRequest = delegationInfo;
+    const delegation: ApiDelegation = {
+      apiName,
+      orgName,
+    };
+    const altinnPartyId =
+      getCookie('AltinnPartyId') === null ? getCookie('AltinnPartyId') : '50067798';
```
Albertlarsen, line 45: This is opposite of what it should be. Don't default to 50067798 when not cookie is null, throw Error instead.

src/components/apiDelegation/given/ReceiptPage/ReceiptPage.tsx
```diff
@@ -15,14 +15,15 @@ import { resetDelegableOrgs } from '@/rtk/features/delegableOrg/delegableOrgSlic
 import { resetDelegableApis } from '@/rtk/features/delegableApi/delegableApiSlice';
 import { RouterPath } from '@/routes/Router';
 import { ReactComponent as ApiIcon } from '@/assets/ShakeHands.svg';
+import { setLoading as setOveviewToReload } from '@/rtk/features/overviewOrg/overviewOrgSlice';
```
Albertlarsen, line 18: Delete the 'as setOveviewToReload', there's no other setLoading here.
```suggestion
import { setLoading } from '@/rtk/features/overviewOrg/overviewOrgSlice';
```

src/rtk/features/delegationRequest/delegationRequestSlice.ts
```diff
@@ -1,86 +1,111 @@
...
+            ],
+          },
+        ],
+      })
+      .then((response) => {
+        return delegation;
+      })
       .catch((error) => {
-        console.log(error);
+        throw error;
```
Albertlarsen, line 69:
```suggestion
      throw new Error(String(error.response.status));
```

src/rtk/features/overviewOrg/overviewOrgSlice.ts
Albertlarsen: Nevermin I can fix this in my branch. Going to nemae to offered.

src/rtk/features/delegationRequest/delegationRequestSlice.ts
allinox: Isn't this how we've had it for a while because of the whole httpOnly thing? 😅

But now that we can just turn that off in the browser loacally, I guess we can remove the default-values we've litterede around in our code - I can do that in this PR 🙂

src/components/apiDelegation/given/ReceiptPage/ReceiptPage.tsx
allinox: Yeah, but I figured it would be more readable like this, since the loading isn't actually for this page, but for another page, which this name makes clear.

src/rtk/features/delegationRequest/delegationRequestSlice.ts
allinox: Won't we lose all the other error information (other than just the status) like this?

I mean, we aren't really using the other info anyway, so it won't be a huge loss I guess, but why take the time to create a new Error when we've already got one we can just pass along? 😅

src/rtk/features/delegationRequest/delegationRequestSlice.ts
Albertlarsen: I don't know. I fixed the issue in my PR in issue 32 so no worries about this comments

Albertlarsen (approved):
Good work 😄

Issue #45, opened 2022-12-02
Create responsive design for S5b | ApiDelegationReceiptPage


Albertlarsen:
This task consists of making component ConfirmationPage for receipt responsive. An example on how it can be made responsive for smartphones can be seen in picture below, but don't follow that example blindly as there most likely will be some errors in the sketch that I haven't spotted yet. Use margins, spacing, etc like you feel makes sense. 

**Additional info:**
We follow the design system's px-definitions on screen sizes: 
- phone view: 280-768 px.
- pc view: 769 px and up

**AC:**
- Page: am.ui.at23.altinn.cloud/accessmanagement/ui/offered-api-delegations/receipt  looks nice on all screen sizes.

Figma link here: https://www.figma.com/file/ESobgAq072y7eQa4gP1JCKER/Profil?node-id=6550%3A84073&t=eAUiGxV6m3Jd5u5b-0

![Image](https://user-images.githubusercontent.com/31471142/212766765-88c9065f-dfaa-4a9f-a6d4-8d72cafa88cb.png)

Randiabs:
sjekket ulike views, ser ok ut. Satt til test

hannekristin:
Satte denne tilbake. Samme som #44 Forslag her: 

![Image](https://user-images.githubusercontent.com/103100644/228819952-96475f69-0877-4b1f-8db6-68ce8f3d9814.png)

Randiabs:
Sjekket, ok. Satt til test

Issue #44, opened 2022-12-02
Create responsive design for S5a | ApiDelegationConfirmationPage


Albertlarsen:
This task consists of making component the page on http://local.altinn.cloud/accessmanagement/ui/api-delegations/confirmation responsive. An example on how it can be made mobile friendly can be seen in picture below, but don't follow that example blindly as there most likely will be some errors in the sketch that I haven't spotted yet. Use margins, spacing, etc like you feel makes sense. 

**Additional info:**
We follow the design system's px-definitions on screen sizes: 
- phone view: 280-768 px.
- pc view: 769px and up

**AC:**
- Page: am.ui.at23.altinn.cloud/accessmanagement/ui/offered-api-delegations/confirmation  looks nice on all screen sizes.

Figma link here: https://www.figma.com/file/ESobgAq072y7eQa4gP1JCKER/Profil?node-id=6550%3A84073&t=eAUiGxV6m3Jd5u5b-0

Figma link here: https://www.figma.com/file/ESobgAq072y7eQa4gP1JCKER/Profil?node-id=6550%3A85281&t=eAUiGxV6m3Jd5u5b-0

![Image](https://user-images.githubusercontent.com/31471142/212767877-614cd4b1-aa66-44eb-b71d-1f4f6d4f726c.png)

hannekristin:
På mobil bør navn på etat og organisasjonsnummer legges på linjen under. Det ser greit ut på iPad mini. 


![Image](https://user-images.githubusercontent.com/103100644/228511288-18fdcc16-0447-435b-bfdf-93448a9874c1.png)


![Image](https://user-images.githubusercontent.com/103100644/228511196-70e96edf-02f9-42b2-b3ff-78cf8ce0f272.png)

Randiabs:
Det ser greit ut i forhold til at etats navnet havner på linjen under. Men har dere fjernet at organisasjonsnummeret for virksomheten som mottar api vises ?  @Albertlarsen

hannekristin:
Org.nr vises heller ikke på større enheter. det gjorde det før siste endring


![Image](https://user-images.githubusercontent.com/103100644/234222259-2865c84b-c475-4733-afcb-302b7a7c6c16.png)

Albertlarsen:
@hannekristin Åja, det skal jeg fikse asap 😊

Albertlarsen:
Jeg kan legge det til men da blir det sånn her på små skjermer:
[...]

PR #225, opened 2023-03-17
Feat/responsive design summary page

Albertlarsen:
## Description
Adds responsive design to receipt Page and confirmation Page

## Related Issue(s)
- #44 
- #45 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/components/reusables/SummaryPage/SummaryPage.tsx
+++ b/src/components/reusables/SummaryPage/SummaryPage.tsx
@@ -43,9 +43,9 @@ export interface SummaryPageProps {
   topListText?: string;
   bottomListText?: string;
   bottomText?: string;
-  nextButtonClick?: () => void;
-  nextButtonDisabled?: boolean;
-  nextButtonLoading?: boolean;
+  confirmationButtonClick?: () => void;
+  confirmationButtonDisabled?: boolean;
+  confirmationButtonLoading?: boolean;
   showNavigationButtons?: boolean;
 }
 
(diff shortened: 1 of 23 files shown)


src/components/reusables/SummaryPage/SummaryPage.tsx
```diff
@@ -34,13 +38,15 @@ export interface SummaryPageProps {
...
   topListText?: string;
   bottomListText?: string;
   bottomText?: string;
-  mainButton?: React.ReactNode;
-  complementaryButton?: React.ReactNode;
-  headerIcon: React.ReactNode;
-  headerColor?: PageColor;
+  nextButtonClick?: () => void;
+  nextButtonDisabled?: boolean;
+  nextButtonLoading?: boolean;
```
allinox, line 48: Why are these props called nextButton when they always refer to a confirmation button?
Can't they be called confirmationButton, so it's more clear that that's what they are? 😅

src/components/reusables/SummaryPage/SummaryPage.tsx
```diff
@@ -34,13 +38,15 @@ export interface SummaryPageProps {
...
+  showNavigationButtons?: boolean;
```
allinox, line 49: Technically, setting this to true still gives you a navigation button, it just takes you back to the overview instead.
I'm not entirely sure what to call it instead though, and it's a nitpick that might just be born from my distaste for the SummaryPage.. 😅

allinox: Yeah, but that's not clear when you just look at the interface of the SummaryPage component by itself.
But like I said, I'm not sure how to fix this other than renaming it to hideBackButton or something, but I'm not sure that's the right way either since the primary button here is a different one depending on this flag. So maybe this is actually a layout-switch or something?

But again, it's a nitpick born from the Frankensteinien premise of the component itself, so I don't think it's worth a lot of effort when I don't even have a good solution to offer.
You can probably just leave it as it is.

Albertlarsen: Yeah I see your point. If there's a lot of issues with working with the component in the future I'll just delete the whole component and separate ConfirmationPage and ReceiptPage more.

PR #262, opened 2023-03-30
Feat/44 responsive design listitem remove middletext

Albertlarsen:
## Description
Fixes stacking of middletext on mobilescreen along with updating DS and adding break-word to ActionBar

## Related Issue(s)
- #44 

## Verification
- [ ] All tests run green

## Documentation


src/components/reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
```diff
@@ -16,7 +16,7 @@ export enum ListTextColor {
 export interface CompactDeletableListItemProps {
   removeCallback?: (() => void) | null;
   leftText: string;
-  middleText: string;
+  middleText?: string | boolean;
```
allinox, line 19: Why a boolean?
Does that mean I can feed the value `true` as a prop and expect a default middle text?

And if not, and this is only meant for the user of the component to be able to indicate they don't want any middle text, can't that just be achived by not passing a string at all? An undefined string is after all evaluated to false anyway.

src/components/reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
```diff
@@ -34,44 +34,88 @@ export const CompactDeletableListItem = ({
...
+          <div className={classes.deleteSection}>
+            {removeCallback && (
+              <Button
+                variant={ButtonVariant.Quiet}
+                color={ButtonColor.Danger}
+                icon={<MinusCircle />}
+                onClick={removeCallback}
+                aria-label={String(t('api_delegation.delete'))}
+              >
+                {!isSm && t('api_delegation.delete')}
```
allinox, line 72: This check of !Sm happens inside another check for Sm so won't it always be false?

src/components/reusables/CompactDeletableListItem/CompactDeletableListItem.tsx
```diff
@@ -34,44 +34,88 @@ export const CompactDeletableListItem = ({
...
+              </Button>
+            )}
+          </div>
+        </div>
+      ) : (
```
allinox, line 77: These two conditional sections have a lot of duplicate code. If checking for the presence of middle text is the only difference between the two, then can't the check for a small screen happen only there?

src/components/reusables/CompactDeletableListItem/CompactDeletableListItem.module.css
```diff
@@ -1,45 +1,100 @@
...
-  margin-right: 10px;
+  .listItemTexts{
+    flex: 1;
+    display: flex;
+    font-size: 16px;
+    --list-text-color-error: #E02E49;
+    align-items: center;
+  }
+
+  .baseListItemContent{
```
allinox, line 59: Looks like there's a lot of duplicate code between the two screen sizes, like this baseListItemContent for example, which seems to be the exact same in both cases.
Maybe only the code affected by the change in screen size should be defined separately for the two cases?

src/components/reusables/SummaryPage/SummaryPage.tsx
```diff
@@ -1,22 +1,22 @@
-import { Panel, PanelVariant } from '@altinn/altinn-design-system';
+import {
+  Page,
+  PageContent,
+  PageHeader,
```
allinox, line 4: Are we going back to the page from the design system or is this a case of git acting up? 😅

allinox (changes requested):
I haven't actually tested this, so I don't know how it looks visually speaking. 😅 I've just taken a quick look at the code.

src/components/reusables/SummaryPage/SummaryPage.tsx
Albertlarsen: git acting up

allinox (approved):
Nice 😊

PR #280, opened 2023-04-17
Wrong 18next english cookie value fix

Albertlarsen:
## Description
Fixes wrong i18next cookie value for english being set

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/resources/LoadLocalizations/LoadLocalizations.tsx
+++ b/src/resources/LoadLocalizations/LoadLocalizations.tsx
@@ -9,7 +9,18 @@ interface Props {
 }
 
 const LoadLocalizations = ({ children }: Props) => {
-  const backupLang = getCookie('i18next') ? getCookie('i18next') : 'no_nb';
+  const getBackupLanguage = () => {
+    const value = getCookie('i18next');
+    if (value === 'no_nn') {
+      return 'no_nn';
+    } else if (value === 'en') {
+      return 'en';
+    } else {
+      return 'no_nb';
+    }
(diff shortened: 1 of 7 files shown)


src/resources/LoadLocalizations/LoadLocalizations.tsx
```diff
@@ -9,7 +9,18 @@ interface Props {
 }
 
 const LoadLocalizations = ({ children }: Props) => {
-  const backupLang = getCookie('i18next') ? getCookie('i18next') : 'no_nb';
+  const getBackupLanguage = () => {
+    const value = getCookie('i18next');
+    if (value === 'no_nn') {
```
acn-dgopa, line 14: Just trying to understand this code block. Are you trying to return a default language? Doesn't BFF already defaults language to bokmål?

allinox (approved):
Good catch!

src/resources/LoadLocalizations/LoadLocalizations.tsx
Albertlarsen: It's meant as a second verification for i18next cookie corresponding with actual localization file names

acn-dgopa: But you are returning the same text as the text in "if" condition.

Albertlarsen: Yeah that was redundant, changed it to just return value now

Issue #272, opened 2023-04-05
Fix hot module replacement not always working


Albertlarsen:
### Description

I have noticed a circular dependency issue related to hot reloading in our project. When we make changes to certain modules, the hot reloading feature is not working properly, and it results in a circular dependency issue. It can be hard to find out where the problem lies but try using diagram tools or visual studio code dependency diagram to find out where the circular dependency is.

Steps to reproduce:

- Run the project with hot reloading enabled.
- Make changes to a module that has a circular dependency.
- Observe that the hot reloading feature is not working, and that there will be an error in the browser console.

### Acceptance Criterias

- The circular dependency issue should be identified and fixed.
- The hot reloading feature should work properly even when there are circular dependencies between modules.
- A test case should be added to ensure that the circular dependency issue does not occur in the future.
- The fix should be merged into the main branch and deployed to the production environment.

PR #288, opened 2023-04-19
Feat/272 fix circular dependency

Albertlarsen:
## Description
Adds some packages checking for circular dependencies

## Related Issue(s)
- #272 

## Verification
- [ ] All tests run green

## Documentation


src/routes/Router/RouterPath.tsx
```diff
@@ -0,0 +1,13 @@
...
+  OfferedApiDelegations = 'offered-api-delegations',
+  Overview = 'overview',
+  ChooseApi = 'choose-api',
+  ChooseOrg = 'choose-org',
+  Confirmation = 'confirmation',
+  Receipt = 'receipt',
+  ReceivedApiDelegations = 'received-api-delegations',
+  Profile = 'Profile',
+  BasePath = '/accessmanagement/ui',
+  ProfileAT23 = 'https://at23.altinn.cloud/ui/Profile',
```
jonkjetiloye, line 11: How does the frontend/client know which environment it's running in to select either ProfileAT23 or ProfileProd (and all the other environments it will be deployed to)?

jonkjetiloye: Could it instead take it's own current domain: https://am.ui.{env}.altinn.cloud/
and just drop the am.ui subdomain?

Albertlarsen: Not sure how that would play out. Profile, is on a completely different domain than the frontend since it's still only in SBL. It is dropping the am.ui subdomain right now.

jonkjetiloye: The domain for TT02 is: `https://tt02.altinn.no/ui/Profile` 
it will be the same across all environments that we'll need to just drop the am.ui subdomain. So if that's whats already is being done I guess the environment specific values here is not needed?

Otherwise the BFF will always know it's deployed environment through environment deploy variables, so the client could receive this info on first initialization perhaps

Albertlarsen: Oh yeah, now I understand what you mean. I'll try it!

src/components/reusables/PageContainer/PageContainer.tsx
```diff
@@ -17,14 +17,7 @@ export const PageContainer = ({ children }: PageContainerProps) => {
...
   const redirectToProfile = () => {
-    const currentEnv = window.location.href;
-    if (currentEnv.includes('am.ui.at23')) {
-      window.location.href = RouterPath.ProfileAT23;
-    } else if (currentEnv.includes('altinn.no')) {
-      window.location.href = RouterPath.ProfileProd;
-    } else {
-      window.location.pathname = '/ui/' + RouterPath.Profile;
-    }
+    window.location.pathname = RouterPath.Profile;
```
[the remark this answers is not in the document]
Albertlarsen: Yeah you're right, I updated the code

jonkjetiloye (approved):

Issue #230, opened 2023-03-21
Create component ActionBar


Albertlarsen:
### Description

**Description**
You are tasked with creating a React component called ActionBar.

Copy the accordion component from digdir design system and make some adjustments to the api and design to comply with picture below.

Tasks: 

- [x] Change name from Accordion to ActionBar on everything
- [x] Add colors for ActionBarHeader
- [x] Implement colors for ActionBarContent
- [x] Style component
- [x] Adjust chevron margins
- [x] Change chevron rotation
- [x] Add hover styling
- [x] Redo title prop
- [x] Adjust vertical margins for HeaderTexts with title and without subtitle
- [x] Redo design after discussion
- [x] Change title API
- [x] Add middle text prop
- [x] Make chevron hideable
- [x] Unify separate components
- [x] Add mouse clicker symbol on hover when ActionBar is expandable
- [x] Make action-buttons take up as little height as possible
- [x] Replace non-expandable ActionBar component in ChooseOrgPage
- [x] Replace Accordions in OverviewPage
- [x] Write tests
- [x] Make ChooseApiPage use DelegationActionBar.

**AC**
- There exists cypress tests testing the component
- It should not be expandable when ActionBar doesn't have children
- Should not show v-symbol when AccordionContent isn't set.
- Should be wcag compliant
- Should have different colors
- Should comply with design in picture below
- Should make css line break for long continous words

**Design review**
- Se på alle sidene som brukte accodion om de nye ActionBarene tilfredstiller designet.

![Image](https://user-images.githubusercontent.com/31471142/233661700-2439b7f5-7c2d-47f6-a490-ecac5d5a9f62.png)

### Additional Information
- Figmalenke: https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?node-id=6632-30965&t=I2supNBdx8fKxsvq-0

_No response_

hannekristin:
ActionBar mangler kant på høyresiden i den grå varianten. Dette gjelder bare på den siden der du legger til API. Ellers ser det veldig bra ut.
![image](https://github.com/Altinn/altinn-access-management-frontend/assets/103100644/1f359752-4c4a-4194-903e-21f58cd2fc42)

Albertlarsen:
Fant en bug: 

![Image](https://github.com/Altinn/altinn-access-management-frontend/assets/31471142/2cbe5552-9625-4359-8cb9-053194333a31)

[...]

PR #356, opened 2023-05-22
Bugfix/230 actionbar dr fixes

Albertlarsen:
## Description
adds design review fixes for ActionBar component

## Related Issue(s)
- #230 

## Verification
- [ ] All tests run green

## Documentation


src/components/DelegationActionBar/DelegationActionBar.tsx
```diff
@@ -0,0 +1,115 @@
...
+          setOpen(!open);
+        }}
+      >
+        {' '}
+        <div className={classes.newApiAccordionContent}>
+          {textList.length > 0 && (
+            <div>
+              <p className={classes.scopeText}>{t('api_delegation.scopes')}:</p>
+              <List borderStyle='dashed'>
+                {textList.map((scope, i) => {
```
allinox, line 77: I'm not in love with the name textList when it pertains to scopes in a component that is this specialized anyway, but if we wish to keep the name as generalized as textList, we should probably stick to that when referring to the list items in the map function.

That being said, I actually think it is more readable when referring to these as scopes, so my favored solution would be to change textList into scopeList. The most important though, is that we are consistant.

allinox (changes requested):

src/components/DelegationActionBar/DelegationActionBar.module.css
```diff
@@ -24,8 +26,12 @@
...
+.actionBarHeaderTitle{
   font-weight: 500;
   margin-top: 2px;
   margin-bottom: 5px;
-}
\ No newline at end of file
+}
+
+.delegationActionBar{
+  margin-right: 10px;
```
allinox, line 36: This is very much a nitpick, but I would probably put the extra space as a padding in the container that has the overflow and scroll bar, instead of a margin on every individual delegationActionBar. Feels more neat I think, since the scroll bar is the reason we're adding this space in the first place.

src/rtk/features/apiDelegation/delegableApi/delegableApiSlice.ts
```diff
@@ -184,6 +184,72 @@ const delegableApiSlice = createSlice({
             }
           }
         }
+        for (let i = 0; i < dataArray.length; i++) {
```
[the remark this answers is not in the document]
Albertlarsen: Oops, needed to replicate some ActionBars to reproduce the safari-bug

src/components/DelegationActionBar/DelegationActionBar.module.css
Albertlarsen: Fixed

src/components/DelegationActionBar/DelegationActionBar.tsx
Albertlarsen: fixed

allinox (approved):
Nice work 😊

Issue #366, opened 2023-05-25
Non-expandable ActionBar should not have focus


allinox:
### Description of the bug

When an ActionBar is not expandable and has no onClick function it should not be focusable as a button.
The actions, however, should be (as they are today).

### Steps To Reproduce

1. Go to Choose Org page
2. Use tab to navigate the page
3. Observe that the ActionBar gets focus even if it is unclickable

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/8e5f3db1-0a4b-4420-8f2a-be1d2a58498a)

Randiabs:
Sjekket, ser ok ut. Satt til test

PR #380, opened 2023-05-31
ActionBar header is now only a button if it has an onClick function

allinox:
## Description

ActionBar header is now only a button if it has an onClick function
When there is no onClick, the header does not get focus, as is the standard of non-action elements.

## Related Issue(s)
- #366 

## Verification
- [ ] All tests run green

## Documentation


Albertlarsen (approved):
Good 💩!

Issue #364, opened 2023-05-25
Focus state on ActionBar looks kinda buggy


allinox:
### Description of the bug

This doesn't actually break any accessibility rules, but it looks kinda weird. It should at least show all the sides of the border and probably also encompass the arrow-down icon to the left. Whether it should outline the entire bar (including the actions) is a question that should probably be discussed with UX.

We should also look into unifying the look of all our focus states so that they look and act similarly to that of the Design system which has recently changed theirs into a black and yellow outline.

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/62b4cab9-f0a3-4362-bceb-bc27c5af85b4)

### Steps To Reproduce

1. go to Overvew page
2. use tab to focus on an ActionBar

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/67594a56-9ef3-4972-8c60-eab2491032f2)

Randiabs:
Sjekket, syns det ser bra ut nå som også nedover pilen er med i fokuset.   Er dette knapp og fokus fra design systemet @allinox og @Albertlarsen  ?

Albertlarsen:
@Randiabs Det er basert på fokus fra design systemet ja

PR #398, opened 2023-06-07
Bugfix/364 add focus state to actionbar

Albertlarsen:
## Description
Adds focus state on action bar when it has children

## Related Issue(s)
- #364 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/components/ActionBar/ActionBar.tsx
+++ b/src/components/ActionBar/ActionBar.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
 /* eslint-disable jsx-a11y/no-static-element-interactions */
 /* eslint-disable jsx-a11y/click-events-have-key-events */
 import * as React from 'react';
@@ -52,7 +53,7 @@ export const ActionBar = ({
             [classes.open]: open,
             [classes.clickable]: onClick,
           })}
-          tabIndex={-1}
+          tabIndex={children ? 0 : -1}
         >
           {children && (
             <div className={cn(classes.actionBarIcon)}>
@@ -68,6 +69,7 @@ export const ActionBar = ({
               data-testid='action-bar'
               aria-expanded={open ?? undefined}
               aria-controls={children ? contentId : undefined}
+              tabIndex={-1}
             >
               <div className={classes.actionBarTexts}>
                 <div className={classes.title}>{title}</div>
--- a/src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
+++ b/src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
@@ -81,7 +81,7 @@ export const ChooseApiPage = () => {
     value: provider,
   }));
 
-  const delegableApiAccordions = () => {
+  const delegableApiActionBars = () => {
     if (error.message) {
       return (
         <ErrorPanel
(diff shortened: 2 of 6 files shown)


src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -102,21 +102,17 @@ export const ChooseApiPage = () => {
...
-            topContentText={api.rightDescription}
-            bottomContentText={api.description}
-            scopeList={api.scopes}
-            buttonType={'add'}
-            onActionButtonClick={() => dispatch(softAddApi(api))}
-            color={'neutral'}
-          ></DelegationActionBar>
-        </div>
+        <DelegationActionBar
+          key={index}
```
allinox, line 106: Should probably use api.id as key instead of index to avoid buggy behaviour

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -128,12 +124,12 @@ export const ChooseApiPage = () => {
         key={index}
       >
         <DelegationActionBar
+          key={index}
```
allinox, line 127: Didn't see this before, but we should probably use api.id here too

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -128,12 +124,12 @@ export const ChooseApiPage = () => {
         key={index}
```
allinox, line 124: Question: Why remove the actionBarWrapper div on the delegatabe apis but not on the chosen ones?

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -212,12 +208,16 @@ export const ChooseApiPage = () => {
...
+                <div className={classes.delegableApisContainer}>
+                  <div className={classes.actionBarWrapper}>{delegableApiActionBars()}</div>
+                </div>
               </div>
               {!isSm && (
                 <div className={common.apiAccordions}>
                   <h4 className={classes.explanationTexts}>{t('api_delegation.chosen_apis')}</h4>
-                  <div className={classes.delegableApisContainer}>{chosenApiAccordions}</div>
+                  <div className={classes.delegableApisContainer}>
+                    <div className={classes.actionBarWrapper}>{chosenApiAccordions}</div>
```
allinox, line 219: Should probably rename this one too
```suggestion
                    <div className={classes.actionBarWrapper}>{chosenApiActionBars}</div>
```

allinox (changes requested):
Haven't tested this locally yet but I'll get to that asap 😊

src/components/ActionBar/ActionBar.tsx
allinox, line 1: I just tested locally and it looks really nice, but I'm no longer able to open it using the keyboard
image

Seems like this area here is the only one one that's clickable, and with the new changes to tab-index we never reach it. But it's strange though... Didn't the area around the action and chevron use to be clickable? 😕
image

allinox: I have a suggestion.

Remove the tab-index adjustments and the entire .actionBar.clickable:focus class and .actionBarHeader:focus (with content) in module.

Then move the chevron icon into the header (into the button) and set `composes: focusable from '@/resources/utils/utility.module.css';` inside .actionBarHeader.

You'll need to fix the padding/margin around the chevron, but it'll look something like this and it will be fully functioning with the keyboard.

image

It's does perhaps not cover the entire bar, but at least it makes it really visible what functionality you're activating if you click.

allinox: And if you want, you can even add a border-radius to the header to make the outline follow the rounded edges either on all or on just the left edges 😊

image

Albertlarsen: I just implemented for sharp edges now (since in reality it is a sharp edge between +-button and header) and it looks a little weird with round on left and sharp on right. We can maybe start with sharp edges and then move to rounded if it's needed later on?

allinox: That's ok for me. And if UX has a problem with it, I'm sure we'll hear it in the design review 👍

allinox (changes requested):
Not done yet but I gotta run so I figured I'd just leave this for now 😅

src/components/ActionBar/ActionBarIcon.tsx
```diff
@@ -19,6 +19,7 @@ export const ActionBarIcon = () => {
     className: cn(iconClassnames),
     'data-testid': 'action-bar-icon',
     onClick,
+    altText: 'chevron',
```
allinox, line 22: I suggest using an aria-hidden on this icon since the word 'chevron' doesn't actually mean a whole lot to people anyway (most people probably don't now what it is)
If you still want an altText, I would suggest 'arrow-down' or something that would mean more to people.

allinox (commented):
Ok, I've gone through it now.
Only one additional suggestion 😊

Issue #376, opened 2023-05-30
Add aria-label text on all SVGs in code base


Albertlarsen:
### Description

Description:
As part of improving accessibility in our code base, we need to ensure that all SVGs used throughout the application have appropriate alt-text. Alt-text provides a textual description of the content of an image for users who may have difficulty seeing or cannot see the image.

Currently, some SVGs in the code base do not have alt-text specified, which can hinder users with visual impairments from understanding the content conveyed by these images. To address this issue, we should add alt-text to all SVGs to ensure a more inclusive and accessible user experience.

Tasks:

Review all SVG files in the code base.
Identify SVGs that are missing alt-text.
Add descriptive alt-text to each SVG, providing a clear and concise description of the image content.
Update the code references to include the alt attribute with the corresponding alt-text.
Expected Outcome:
By adding alt-text to all SVGs, we will improve the accessibility of our application, making it more inclusive and user-friendly for individuals with visual impairments. This enhancement aligns with our commitment to accessibility best practices and ensures compliance with accessibility standards.

Note: It's important to consider the context and content of each SVG when writing alt-text to accurately convey the meaning or purpose of the image.

allinox:
@Albertlarsen 

Ref comment on PR 399 - are you working on this right now?

Albertlarsen:
@allinox nope, I'm not working on anything that isn't in 'in progress'

allinox:
@Albertlarsen 

Ok, I was just wondering since you commented about the topic on another PR a couple of days ago 👍

Albertlarsen:
@allinox 
this task is perhaps not that relevant after our discussion. Also it seems like alt-texts is not needed on icons that's just for illustrations.

allinox:
@Albertlarsen 
True, if the goal is accessibility, we're already covered with aria-labels in that regard, so the only thing it would help for are the cases with the add and delete buttons if the user has poor connection so the image isn't loaded. But I don't know how far we should stretch ourselves to cater to cases like that.

PR #399, opened 2023-06-07
removed aria-label and added aria-hidden to filter button chevron

allinox:
## Description
Screen readers read the chevron out loud as open or closed when reading the label for the filer button, something that didn't quite work when the filter button is also used to label the dialog box it opens

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/components/Filter/FilterButton.tsx
+++ b/src/components/Filter/FilterButton.tsx
@@ -71,7 +71,7 @@ export const FilterButton = forwardRef<HTMLButtonElement, FilterButtonProps>(
           <SvgIcon
             svgIconComponent={<ChevronDownIcon />}
             className={cn(classes.icon, classes.chevron, { [classes.open]: isOpen })}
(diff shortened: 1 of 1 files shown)


Albertlarsen (approved):
In this case I agree that it's better to have aria-hidden since the button says 'Filtrer på etat' so it's unnecessary with another text saying 'filter'. The following is irrelevant for this PR but since I'm not on the office i use this channel to discuss it :p In most cases I think we should have altText, see illustration below.

image

allinox:
@Albertlarsen 

I agree, except with the chevron. I don't think it provides a whole lot of nessecary information anyway other than indicating that this element is expandable (which awe have indicated in other way's anyway) so I think it could be aria-hidden as well. In fact, this is what the Designsystem has done with their accordion.

There is a task #376 that could perhaps be used to further this discussion and to decide if this should be a prioritized task atm

Issue #397, opened 2023-06-07
Implement new close button to Improve WCAG 3.0 Compliance


Albertlarsen:
### Description

Description:
We need to implement a new close button that meets Web Content Accessibility Guidelines (WCAG) standards. The current close button does not provide sufficient accessibility features, and we want to ensure that all users, including those with disabilities, can easily interact with it.

Tasks:

Replace existing close button:

1. Identify the existing close button element/component in the codebase.
2. Replace the existing close button with an accessible alternative that meets WCAG guidelines.
3. Add appropriate accessibility attributes:
Ensure the new close button includes the appropriate accessibility attributes, such as aria-label, role, and tabindex, to provide meaningful information to assistive technologies.
Implement keyboard accessibility:

4. Verify that the new close button can be easily accessed and triggered using keyboard navigation alone.
Test the button's functionality using keyboard-only interactions and make any necessary adjustments to ensure smooth accessibility.
Test with assistive technologies:

5. Validate the new close button's behavior and accessibility using screen readers, such as NVDA, VoiceOver, or JAWS.
Identify and address any issues or discrepancies found during testing.

The button can be found here: https://www.figma.com/file/ESobgAq072y7eQa4gP1JCKER/Profil?type=design&node-id=8025-225850&t=fFB01mPqM8TKNHFE-0

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/31471142/a3cd6d88-8c07-414f-adeb-cee50c74f937)

**Design review:**

- Gå til api-delegering og test ut lukk-knappen

hannekristin:
Testet ok

PR #410, opened 2023-06-08
Redesign close button

Albertlarsen:
## Description
This PR adds new wcag 3-compliant colors to the close button

## Related Issue(s)
- #397 

## Verification
- [ ] All tests run green

## Documentation


src/components/PageContainer/PageContainer.tsx
```diff
@@ -2,7 +2,7 @@ import { Button, ButtonColor, ButtonSize, ButtonVariant } from '@digdir/design-s
 import * as React from 'react';
```
allinox, line 2: The line above this has a lot of warnings 😅
Now that DS Button is no longer used here, the imports for it can be removed.

src/assets/CloseIcon.png
allinox, line 1: Do we need this asset if only the svg is used?

src/components/PageContainer/PageContainer.module.css
```diff
@@ -29,4 +29,20 @@
...
   justify-content: flex-end;
+}
+
+.closeButton{
+  border-radius: 50%;
+  background-color: #FFFFFF;
+  border: none;
+  padding: 0px;
+  height: 36px;
+  padding-top: 3px;
```
allinox, line 40: Why the mix of padding and margin?
Why not just stick to one?

src/components/PageContainer/PageContainer.tsx
```diff
@@ -26,14 +26,16 @@ export const PageContainer = ({ children }: PageContainerProps) => {
       <div className={classes.pageContainer}>
         <UserInfoBar />
```
allinox, line 27: The page looks a little bit cramped with the UserInfo and the Exit button so close together imo. We probably don't need as much space between them as in the design, but I do think we need a bit more than this.
Maybe a gap of 30px or something?

image

allinox (changes requested):
Nice! I like the look of this and we're finally WCAG compliant! 😊
Just a few nitpicks here really

src/components/PageContainer/PageContainer.module.css
Albertlarsen: They do different stuff and if I don't use both the x will not be centered. It's a result of trial and error, not any calculations. It's a little bit messy and if this causes problems for us in future we should probably make UX draw up a red x-icon that fits this size of button, with the exactly the right viewbox to fit this size of button (as this is something that also affects the centering of the x-icon). I just thought this was a static button that's only going to be used in this exact size and we probably won't adjust the sizes later on.

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -1,5 +1,5 @@
 import { PanelVariant, Panel, SearchField } from '@altinn/altinn-design-system';
-import { List, Spinner } from '@digdir/design-system-react';
+import { List, Spinner, Alert } from '@digdir/design-system-react';
```
allinox, line 2: Is the Alert used here?

src/components/PageContainer/PageContainer.module.css
```diff
@@ -16,6 +16,10 @@
   }
 }
 
+.pageMargin{
+  margin-top: 30px;
```
allinox, line 20:
```suggestion
  margin-top: 20px;
```
Just a suggestion to save some space on mobile 😊

allinox (approved):
Nice! 😄
Just a small suggestion and a tiny question, but otherwise.. APPROVED! 🎉

Issue #391, opened 2023-06-06
Sette opp ny route til Enkeltrettigheter


allinox:
### Description

Denne oppgaven handler om å forberede routing til Enkeltrettigheter.
Den er todelt:

- Lag ny enum i Router-paths for bruk til Enkeltrettigheter og omdøp den allerede eksisterende enumen til å infikere dens bruk i ApiDelegering (Forslag: ApiDelegationRoute)
- Undersøk om det er gunstig å ha flere enn én router (en per hovedfeature). Hvis det er det, sett opp for ulike på Api-delegering og Enkeltrettigheter. (Obs: Test antall renders som skjer or routing-element)

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/d8cc8622-65a8-4c85-ad8c-0f1b425e1f37)

### Design review
Sjekk at alle flytene og navigeringsknapper fungerer som de skal

Albertlarsen:
In a typical React application, it is generally recommended to have a single Router component that handles the routing for the entire application. This is because the Router is responsible for managing the application's URL and mapping it to different components or views.

Having multiple Router components can lead to conflicts and unexpected behavior, as each Router maintains its own routing state. It can result in difficulties in managing nested routes, maintaining consistent URL synchronization, and coordinating navigation across different parts of the application.

However, there might be specific scenarios where you may have a valid use case for multiple routers. For example, if you have a large application with distinct sections that are completely separate from each other and don't need to share routing information, you could potentially use multiple routers.

But in most cases, a single top-level Router, such as the BrowserRouter or HashRouter from React Router, is sufficient to handle routing for the entire application. It provides a centralized and consistent way to manage navigation and URL changes across different components and views.

Remember to design your application's routing structure carefully and consider the overall architecture to determine the appropriate usage of routers.

Randiabs:
sjekket ok

PR #421, opened 2023-06-13
Feat/single rights/391 routes for singlerights

Albertlarsen:
## Description
Adds a route and a dummy page for single rights overview

## Related Issue(s)
- #391 

## Verification
- [ ] All tests run green

## Documentation


src/features/apiDelegation/components/OverviewPageContent/OverviewPageContent.tsx
```diff
@@ -88,7 +89,8 @@ export const OverviewPageContent = ({
 
   const goToStartDelegation = () => {
     dispatch(restoreAllSoftDeletedItems());
-    navigate('/' + RouterPath.OfferedApiDelegations + '/' + RouterPath.ChooseOrg);
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
```
allinox, line 92: Why this disabling?
What is it complaining about that it wasn't before?
And can it be solved some other way?

src/features/singleRight/offered/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,20 @@
+import { useTranslation } from 'react-i18next';
+import * as React from 'react';
+import { PersonCheckmarkIcon } from '@navikt/aksel-icons';
+
+import { Page, PageHeader, PageContent, PageSize, PageContainer } from '@/components';
+import { useMediaQuery } from '@/resources/hooks';
+
+export const OverviewPage = () => {
+  const { t } = useTranslation('common');
```
allinox, line 9: I get an assigned but never used warning on this one.

src/routes/paths/ApiDelegationPath.tsx
```diff
@@ -7,5 +7,4 @@ export enum RouterPath {
   Receipt = 'receipt',
   ReceivedApiDelegations = 'received-api-delegations',
   Profile = 'ui/Profile',
```
allinox, line 9: Should Profile perhaps be in GeneralPaths?

src/routes/Router/Router.tsx
```diff
@@ -19,36 +22,52 @@ export const Router = createBrowserRouter(
       errorElement={<NotFoundSite />}
     >
       <Route
-        path={RouterPath.OfferedApiDelegations + '/' + RouterPath.Overview}
+        path={ApiDelegationPath.OfferedApiDelegations}
```
allinox, line 25: Is this a path? Just the OfferedApiDelegations, without the overview?

src/routes/Router/Router.tsx
```diff
@@ -19,36 +22,52 @@ export const Router = createBrowserRouter(
...
         element={<OfferedOverviewPage />}
         errorElement={<NotFoundSite />}
-      />
+      >
+        <Route
+          path={ApiDelegationPath.Overview}
```
allinox, line 30: Shouldn't they follow this pattern?

```suggestion
          path={ApiDelegationPath.OfferedApiDelegations + '/' + ApiDelegationPath.Overview}
```

src/routes/paths/SingleRightPath.tsx
```diff
@@ -0,0 +1,4 @@
+export enum SingleRightPath {
+  OfferedSingleRights = 'offered-single-rights',
```
allinox, line 2: Should this be offered? Perhaps DelegateSingleRights is better?
```suggestion
  DelegateSingleRights = 'delegate-single-rights',
```

src/routes/Router/Router.tsx
allinox, line 1: I tried running this locally and I get stuck on the overviewPage no matter what path I try (like receipt and choose-org)
Hitting the 'delegate new API'-button doesn't take me anywhere new either, only resulting in an error: `Uncaught ReferenceError: RouterPath is not defined`

I suspect it is because of the setup in this file as I have commented on above)

image

src/routes/Router/Router.tsx
```diff
@@ -9,8 +9,11 @@ import { ChooseOrgPage } from '@/features/apiDelegation/offered/ChooseOrgPage';
 import { ReceiptPage } from '@/features/apiDelegation/offered/ReceiptPage';
 import { ConfirmationPage } from '@/features/apiDelegation/offered/ConfirmationPage';
 import { NotFoundSite } from '@/sites/NotFoundSite';
+import { OverviewPage } from '@/features/singleRight/offered/OverviewPage/OverviewPage';
```
allinox, line 12:
```suggestion
import { OverviewPage as SingleRightsOveview } from '@/features/singleRight/offered/OverviewPage/OverviewPage';
```
That being said.. Are we even gonna have an overview page on single rights? 🤔

src/features/singleRight/offered/OverviewPage/OverviewPage.tsx
allinox, line 1: Don't know where to put a comment like this, cause it's really more on the folder structure than this specific page (Although, maybe this shouldn't be named overview either now that I think about it, since we're not planning on having an overview.. But I assume this is just temporary and will be changed later anyway)

_Anyway_, what I was actually gonna comment on (before I distracted myself 😅) was the folder name 'offered'. 
I don't think there will be an offered/received divide here. Instead, I think the flows will be delegate, request, and approve. So maybe rename the offered-folder into 'delegate'?

allinox (changes requested):
Finally we're starting on the singleRights delegation! Awesome!😄

src/features/apiDelegation/components/OverviewPageContent/OverviewPageContent.tsx
Albertlarsen: Ahh I forgot to change the old RouterPath.ChooseOrg to ApiDelegationPath.ChooseOrg

src/features/singleRight/offered/OverviewPage/OverviewPage.tsx
Albertlarsen: True!

src/features/singleRight/offered/OverviewPage/OverviewPage.tsx
Albertlarsen: I'll remove it

src/routes/Router/Router.tsx
Albertlarsen: Nice catch, I had to change something in the Router to make this work. 

Appearantly it's not possible to have the parent Route with an element

src/routes/Router/Router.tsx
Albertlarsen: Don't know what to call it yet, I thought the first page could be called overview. Maybe I should change it to StartPage? Open for suggestions!

src/routes/Router/Router.tsx
Albertlarsen: Its a way of structuring different child routes to parent routes to avoid errors and duplication of code. See the new Router I'll soon push for more like what it should be like

src/routes/paths/ApiDelegationPath.tsx
Albertlarsen: Yes!

src/routes/Router/Router.tsx
allinox: We start right on the delegation, don't we?
Maybe chooseService? And then the next one we can call specifyRights?

src/routes/Router/Router.tsx
Albertlarsen: With that change we also got hot-reload on direct elements 🥳 
(It's probably the correct way of doing it)

allinox: Aaaahh, I didn't realize that they were parent and child with the diff being all over the place
But that makes sense! 😁

src/components/PageContainer/PageContainer.tsx
```diff
@@ -17,7 +18,8 @@ export const PageContainer = ({ children }: PageContainerProps) => {
 
   const redirectToProfile = () => {
     const cleanHostname = window.location.hostname.replace('am.ui.', '');
-    window.location.href = 'https://' + cleanHostname + '/' + ApiDelegationPath.Profile;
+    window.location.href =
+      'https://' + cleanHostname + '/' + GeneralPath.Profile + '?R=' + getCookie('AltinnPartyId');
```
allinox, line 22: I'm assuming this is to keep the user as the same party (org) when sending them back to the profile?
(Hard to test this locally so I'm just asking 😅)

allinox (approved):
Great work! 🙌🎉

src/components/PageContainer/PageContainer.tsx
Albertlarsen: Yeah it is because there's a bug in Altinn2

PR #455, opened 2023-07-04
Feat/single rights/393 bff implement search api

allinox:
## Description

Created BFF api for paginated search that fetches all (non maskinportenschema) resources from the resource registry and returns those matching the given filters and search string if given.

The code can be tested locally by setting the useMockData value in appsettings.Development and running the endpoint through swagger (or through postman if you prefer that)

## Related Issue(s)
- #393 

## Verification
- [ ] All tests run green

## Documentation


backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
```diff
@@ -64,6 +65,127 @@ public async Task GetMaskinportenResources_valid()
...
+            List<ServiceResourceFE> allExpectedResources = TestDataUtil.GetSingleRightsResources();
+            PaginatedList<ServiceResourceFE> expectedResult = new PaginatedList<ServiceResourceFE>(allExpectedResources.GetRange(0, numPerPage), page, allExpectedResources.Count);
+
+            // Act
+            HttpResponseMessage response = await _client.GetAsync($"accessmanagement/api/v1/resources/paginatedSearch?NumPerPage={numPerPage}&Page={page}");
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            PaginatedList<ServiceResourceFE> actualResources = JsonSerializer.Deserialize<PaginatedList<ServiceResourceFE>>(await response.Content.ReadAsStringAsync(), options);
```
Albertlarsen, line 90: I think this line should be moved to under // Act or // Arrange
```suggestion
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
```diff
@@ -64,6 +65,127 @@ public async Task GetMaskinportenResources_valid()
...
+            List<ServiceResourceFE> allExpectedResources = TestDataUtil.GetSingleRightsResources().FindAll(r => roFilters.Contains(r.ResourceOwnerName.ToLower()));
+            PaginatedList<ServiceResourceFE> expectedResult = new PaginatedList<ServiceResourceFE>(allExpectedResources.GetRange(0, numPerPage), page, allExpectedResources.Count);
+
+            // Act
+            HttpResponseMessage response = await _client.GetAsync($"accessmanagement/api/v1/resources/paginatedSearch?NumPerPage={numPerPage}&Page={page}&ROFilters={roFilters[0]}&ROFilters={roFilters[1]}");
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            PaginatedList<ServiceResourceFE> actualResources = JsonSerializer.Deserialize<PaginatedList<ServiceResourceFE>>(await response.Content.ReadAsStringAsync(), options);
```
Albertlarsen, line 119: Assert section should only contain Assert methods I think

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Models/ResourceRegistry/PaginatedList.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// The list entries pertaining to the current page
+        /// </summary>
+        public List<T> PageList { get; set; }
+
+        /// <summary>
+        /// Parameterless constructor
+        /// </summary>
+        public PaginatedList()
```
Albertlarsen, line 32: Why do you need the Parameterless constructor?

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
```diff
@@ -64,6 +65,127 @@ public async Task GetMaskinportenResources_valid()
...
+        /// Expected: PaginatedSearch returns a list of all single rights resources in paginated form with language filtered for the authenticated users selected language
+        /// </summary>
+        [Fact]
+        public async Task GetSingleRightsSearch_searchStringAndFilterNotSet_ReturnsAll()
+        {
+            // Arrange
+            var token = PrincipalUtil.GetToken(1337, 501337, 2);
+            _client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", token);
+            int page = 1;
+            int numPerPage = 7;
```
Albertlarsen, line 79:
```suggestion
            int resultsPerPage = 7;
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
```diff
@@ -64,6 +65,127 @@ public async Task GetMaskinportenResources_valid()
...
+        public async Task GetSingleRightsSearch_searchStringSet_ReturnsMatch()
+        {
+            // Arrange
+            var token = PrincipalUtil.GetToken(1337, 501337, 2);
+            _client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", token);
+            int page = 1;
+            int numPerPage = 1;
+            string searchString = "klesskapet";
+
+            List<ServiceResourceFE> allExpectedResources = TestDataUtil.GetSingleRightsResources().FindAll(r => r.Title.ToLower().Contains(searchString));
```
Albertlarsen, line 139: Why don't you reuse the methods you created in the RAP instead of duplicating the same thing rap.SearchInResourceList does on this line?

Albertlarsen (changes requested):
Bra greier! Ikke ferdig med review enda, gjør mer i morgen 😊

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Utils/TestDataUtil.cs
```diff
@@ -63,6 +63,29 @@ public static List<ServiceResourceFE> GetExpectedResources(ResourceType resource
             return filteredResources;
         }
 
+        public static List<ServiceResourceFE> GetSingleRightsResources()
+        {
+            List<ServiceResourceFE> resources = new List<ServiceResourceFE>();
+            List<ServiceResourceFE> filteredResources = null;
+
+            string path = GetResourcesPath("resourcesfe");
```
Albertlarsen, line 71: so here in TestDataUtil the file path is resourcesfe but seems like ResourceRegistryClientMock goes to the resource files. Why do they go to different places?

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
```diff
@@ -64,6 +65,127 @@ public async Task GetMaskinportenResources_valid()
...
+            HttpResponseMessage response = await _client.GetAsync($"accessmanagement/api/v1/resources/paginatedSearch?NumPerPage={numPerPage}&Page={page}&ROFilters={roFilters[0]}&ROFilters={roFilters[1]}");
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            PaginatedList<ServiceResourceFE> actualResources = JsonSerializer.Deserialize<PaginatedList<ServiceResourceFE>>(await response.Content.ReadAsStringAsync(), options);
+            Assert.Equal(expectedResult.Page, actualResources.Page);
+            Assert.Equal(expectedResult.NumEntriesTotal, actualResources.NumEntriesTotal);
+            AssertionUtil.AssertCollections(expectedResult.PageList, actualResources.PageList, AssertionUtil.AssertResourceExternalEqual);
+        }
```
Albertlarsen, line 123: Think it would be nice to also have a test that returns another page than just page 1 as all the tests only really checks page 2

Albertlarsen (changes requested):

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/IResourceAdministrationPoint.cs
```diff
@@ -9,6 +9,12 @@ namespace Altinn.AccessManagement.UI.Core.Services.Interfaces
     /// </summary>
     public interface IResourceAdministrationPoint
     {
+        /// <summary>
+        /// Fetches the singleRights resources, processes them with filters and search string if given and then paginates the result
+        /// </summary>
+        /// <returns>the resources that match the filters and search string corresponding to the provided page</returns>
+        Task<PaginatedList<ServiceResourceFE>> GetPaginatedSearchResults(string languageCode, string[]? resourceOwnerFilters, string? searchString, int page, int numPerPage);
```
Albertlarsen, line 16: When seeing the description for this class and this description when searching online:

> Overall, the Resource Administration Point is responsible for centralizing the management, evaluation, and distribution of access control policies within an access control system. It provides the necessary infrastructure for administrators to define and enforce policies that govern the authorization of resource access.

I'm not really sure if the methods you created, belongs in this class. Maybe we should talk to Jon Kjetil about this

Albertlarsen: But I'm not really sure because I see it already contains a Get method 😅  so yeah we should probably talk to Jon Kjetil tomorrow and get clearified what it this file should contain

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
allinox: Honestly, I just followed the already established convention in the file, but you do have a point that this isn't strictly an Assert.

Though, I'm not sure it fits completely anywhere 😅
It is a reading of the results resulting from an Act so it can't be in Arrange because it happens before Act, but it isn't an Act in itself.

Also, deserializing the results is only possible if the response has a statuscode of OK, thus it makes sense to check for that first so we get an assertion fail rather than a thrown exception and that _is_ in fact an assertion, and... you know what, I've actually argued myself back to thinking this line should remain where it is 😅

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Models/ResourceRegistry/PaginatedList.cs
allinox: Because json deserialization needs one to work (or at least it complained that it did before I added it)

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
allinox: Because then I wouldn't actually be testing whether the code ran as it should and errors may pass as expected results. If I check the code by using the same code, it would be sorta like checking if a nail is the same as a hammer by comparing it to a picture of that same nail I am currently expecting.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
```diff
@@ -64,6 +65,127 @@ public async Task GetMaskinportenResources_valid()
...
+        /// Expected: PaginatedSearch returns a list of resources matching the search string in paginated form with language filtered for the authenticated users selected language
+        /// </summary>
+        [Fact]
+        public async Task GetSingleRightsSearch_searchStringSet_ReturnsMatch()
+        {
+            // Arrange
+            var token = PrincipalUtil.GetToken(1337, 501337, 2);
+            _client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", token);
+            int page = 1;
+            int numPerPage = 1;
```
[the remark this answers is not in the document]
allinox: I'll add a test that checks that the pagination works as expected 👍

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/Controllers/ResourceController.cs
```diff
@@ -46,14 +48,45 @@ public class ResourceController : ControllerBase
...
         }
+
+        /// <summary>
+        /// Search through all delegable service resources and returns matches
+        /// </summary>
+        /// <returns>Paginated search results</returns>
+        [HttpGet]
+        [Authorize]
+        [Route("paginatedSearch")]
+        public async Task<ActionResult<PaginatedList<ServiceResourceFE>>> SearchExtended([FromQuery] PaginatedSearchParams parameters)
```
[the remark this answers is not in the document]
allinox: Good point!
I must have forgotten to change the actual method name when I changed the route 😅

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Utils/TestDataUtil.cs
allinox: That's right. `resources` is the list of all reasources and is in many way's a mock of the resource registry. `resourcefe` however is the same but in the format expected by the frontend, so without the different language alternatives.
In fact, the only difference between the two should be that `resources` contains the resources in the ServiceResource format, while `resourcefe` contains them in the ServiceResourceFE format 😊

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
allinox: Yeah, I'll put that in the previously mentioned test for pagination specifically

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/IResourceAdministrationPoint.cs
allinox: Yeah, I just followed the establiched convention on the file (or at least that was my intention), but I don't know if that convention was right in the first place 😅

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/ResourceControllerTest.cs
Albertlarsen: Yeah it makes sense sequentially and can actually value as we will be able to spot earlier if there are errors related to doing the actual the call. But just looking how it already is in current file or tests isn't necessarily a convention and normally the desirialzing happens in the Act, although it may make sense to have it under Assert as well

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Models/ResourceRegistry/Frontend/PaginatedSearchParams.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// List of Resource Owners to be used for filtering
+        /// </summary>
+        public string[]? ROFilters { get; set; }
+
+        /// <summary>
+        /// Number of returned resources (per page)
+        /// </summary>
+        public int NumPerPage { get; set; }
```
Albertlarsen, line 27:
```suggestion
        public int ResultsPerPage { get; set; }
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Models/ResourceRegistry/PaginatedList.cs
Albertlarsen: But you have another constructor below it. In which file do you get the error?

allinox: Yes, but the other constructor is not parameterless, and apparently that is what deserialization requires. It is the same with ServiceResourceFE where I also added a parameterless constructor because the tests failed when it didn't have one

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/IResourceAdministrationPoint.cs
allinox: Since you're renaming the file in your PR, I'll just leave it as-is in mine

Albertlarsen (approved):
Nice work 🤩

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Helpers/PaginationUtils.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// Paginates a list, returning the list elements pertaining to the given page
+        /// </summary>
+        /// <typeparam name="T">The type contained in the list</typeparam>
+        /// <param name="list">The list to be paginated</param>
+        /// <param name="page">The page number of the returned elements </param>
+        /// <param name="numPerPage">Number of elements on each page</param>
+        /// <returns>The elements present on the given page</returns>
+        /// <exception cref="ArgumentOutOfRangeException">This exception is thown when either numPerage or page is not valid, souch as if asking for a page does not exist</exception>
+        public static PaginatedList<T> GetListPage<T>(List<T> list, int page, int numPerPage)
```
jonkjetiloye, line 24: I would think that it can be safer and perhaps more efficient to just use LINQ for the pagination.
Something like: 
`list.Skip(page*pageSize).Take(pageSize).ToList()`

Have you looked into something like that? Might still need some logic to handle/check argument sizes against the size of the list to avoid out of bounds issues.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/Controllers/ResourceController.cs
```diff
@@ -46,14 +48,45 @@ public class ResourceController : ControllerBase
         /// <returns>List of API service resources</returns>
         [HttpGet]
         [Authorize]
-        [Route("accessmanagement/api/v1/resources/maskinportenschema")]
-        public async Task<ActionResult<List<ServiceResourceFE>>> Get()
+        [Route("maskinportenschema")]
+        public async Task<ActionResult<List<ServiceResourceFE>>> GetMaskinportendchema()
```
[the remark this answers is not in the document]
allinox: Oops 😅
Good catch!

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
```diff
@@ -154,23 +177,110 @@ private async Task<List<ServiceResource>> GetResources()
             return resources;
         }
 
+        private async Task<List<ServiceResource>> GetFullResourceList()
+        {
+            string cacheKey = $"resources:fulllist";
+            if (!_memoryCache.TryGetValue(cacheKey, out List<ServiceResource> resources))
+            {
+                // TODO: Rewrite current GetResurces to use new API instead of search?
```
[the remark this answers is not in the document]
allinox: Well, it was actually a question aimed more at the other GetResources (the one for maskinportenSchemas), since that one currently uses search.
But I think I've decided to just leave it as-is for the time being so this todo should probably be removed

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
```diff
@@ -154,23 +177,110 @@ private async Task<List<ServiceResource>> GetResources()
...
+                {
+                    if (res.Title.ToLower().Contains(word) || res.Description.ToLower().Contains(word) || res.RightDescription.ToLower().Contains(word))
+                    {
+                        numMatches++;
+                    }
+                }
+
+                if (numMatches > 0)
+                {
+                    matchedResources.Add(new SortableServiceResourceFE(res, numMatches));
```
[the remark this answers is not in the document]
allinox: Huh, I guess we could..
I mean, my thought was that I didn't want to mess with an already established and much used model, but if I just make the property nullable I guess it will just go ignored when not applicable 🤔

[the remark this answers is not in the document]
jonkjetiloye: I assume we probably want to move the search/filter implementation for the existing use of the ServiceResourceFE model in the API delegation UI also to use the same implementation in BFF later anyway, and support paging there as well.

Just a thought, but it's not critical at this point. Can be cleaned up when rewriting to support it for APIs later.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Helpers/PaginationUtils.cs
allinox: Not sure what you mean here..
Like instead of GetRange? They basically do the same don't they? 😅

I just did a quick check on StackOverflow and according to the accepted answer there:
"The only difference is that List.GetRange is more efficient than Take(n).ToList() since it already knows the size of the new list whereas the LINQ methods don't know it's size."

jonkjetiloye: Interesting. The StackOverflow aswer is getting a bit old though. LINQ has had a lot of optimization lately.

It was mostly a thought of whether there was an actual need for a PaginatedList model / Util and custom logic if it could be solved by LINQ directly. Otherwise I don't see any issue with your implementation.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
```diff
@@ -37,6 +40,26 @@ public class ResourceAdministrationPoint : IResourceAdministrationPoint
...
+        /// <inheritdoc />
+        public async Task<PaginatedList<ServiceResourceFE>> GetPaginatedSearchResults(string languageCode, string[]? resourceOwnerFilters, string? searchString, int page, int resultsPerPage)
+        {
+            try
+            {
+                List<ServiceResource> resources = await GetFullResourceList();
+                List<ServiceResource> resourceList = resources.FindAll(r => r.ResourceType != ResourceType.MaskinportenSchema);
+                List<ServiceResourceFE> resourcesFE = MapResourceToFrontendModel(resourceList, languageCode);
+
+                List<ServiceResourceFE> filteredresources = FilterResourceList(resourcesFE, resourceOwnerFilters);
```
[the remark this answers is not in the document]
allinox: But isn't that just the same number of operations anyway? 😅
The fact that they're separated into different loops shouldn't really matter, should it?

I mean, you get a performance of `O(n*f)` (where n is total num of resources and f is num of filters) for the filter-loop and then `O(m*s)`  (where m is num of resources and s is num of total search words) for the search, where f<<n, s<<m and m<=n, making the complete worst case scenario `O(n) + O(n)` which is `O(n)` anyway.

I guess the creation of extra variables is a thing, since you do store data in between function calls, but I don't know how much that affects performance in the grand scheme of things. I do think it is much more readable and reusable if we separate the operations though.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
```diff
@@ -154,23 +177,110 @@ private async Task<List<ServiceResource>> GetResources()
             return resources;
         }
 
+        private async Task<List<ServiceResource>> GetFullResourceList()
+        {
+            string cacheKey = $"resources:fulllist";
+            if (!_memoryCache.TryGetValue(cacheKey, out List<ServiceResource> resources))
```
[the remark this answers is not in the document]
allinox: Ok, then I'll remove it 👍

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
```diff
@@ -154,23 +177,110 @@ private async Task<List<ServiceResource>> GetResources()
...
+            {
+                return resources;
+            }
+
+            List<ServiceResourceFE> filteredResources = new List<ServiceResourceFE>();
+            string[] lowercaseResourceOwnerFilters = Array.ConvertAll(resourceOwnerFilters, d => d.ToLower());
+
+            foreach (ServiceResourceFE res in resources)
+            {
+                if (lowercaseResourceOwnerFilters.Contains(res.ResourceOwnerName.ToLower()))
```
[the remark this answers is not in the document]
allinox: I'll change it to using the OrgNumbers instead 😊

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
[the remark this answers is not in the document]
allinox: I've already changed it 😛

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceAdministrationPoint.cs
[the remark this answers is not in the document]
jonkjetiloye: Fair point, and yes the gains might be too minor to make it worth while over having an implementation which is easier to maintain, reuse and expand upon.

We'll anyway know more of how it performs once we get this to TT02/YT and can revisit both code and caching when we see if and where the shoe pinches.

jonkjetiloye (approved):
Awesome sauce! 👍

Issue #495, opened 2023-08-18
API delegering - virksomhet søk gir ikke noe treff


anishastphen:
Gå til API delegering side
Start API delegering og søk etter en org - gir ikke noe treff
Den pleide å funke i går, nå feiler det i alle AT miljøer

<img width="558" alt="image" src="https://github.com/Altinn/altinn-access-management/assets/117180726/b8823967-eb31-4e32-9af4-1ae07443053f">

Albertlarsen:
Please create these issues in altinn-access-management-frontend repo instead so we don't have to transfer it before referring it in our PRs

Albertlarsen:
Testet akkurat fiksen som ble pushet på dette @anishastphen, nå skal det funke igjen i alle andre miljøer enn AT23 😁

allinox:
@anishastphen 
I've added a fix that works in AT23 as well. So that you can search for orgs.

However, since the mock is now active in this environment, you can only search for orgs that are present in the mock data.
To this end I've added 910510789 (VIKESÅ OG ÅLVUNDFJORD) and 991825827 (Digitaliseringsdirektoratet) to the registry mock data so that these are now available for search.
But delegation to them will unfortunately fail.

If there is a certain user (with org) that you usually use for testing and an org you usually use for delegation, you can send me their info (with partyid's) and I can add them to the full flow of the mock data.
Otherwise, the mocking will only be a temporary thing anyway, so maybe it's tolerable that just AT23 will have some failing tests for now.

anishastphen:
fix works in AT22 and AT24, fails only in AT23
so will close the issue for now and will check in AT23 once its available

PR #497, opened 2023-08-18
Fixed faulty org lookup

allinox:
## Description
The error came from the setup of the lookup service, which failed when mocking, not because the lookup service itself was mocked, but because it uses the registry client and that was mocked. The reason for the error was that the lookup service expected the register client to have a constructor with certain parameters that was not made for the mocked version of the register client, as it was not necessary during test. This constructor has now been added.

However, due to the registry being mocked, only orgs present in the mock data can be searched for. For this reason I've added 910510789 (VIKESÅ OG ÅLVUNDFJORD) and 991825827 (Digitaliseringsdirektoratet) to the registry mock data. These are now available for search.

## Related Issue(s)
- #495 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
+++ b/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
@@ -1,11 +1,8 @@
 ﻿using System.Text.Json;
-using System.Text.Json.Serialization;
 using Altinn.AccessManagement.UI.Core.ClientInterfaces;
 using Altinn.AccessManagement.UI.Core.Services.Interfaces;
 using Altinn.Platform.Register.Models;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 
 namespace Altinn.AccessManagement.UI.Mocks.Mocks
 {
--- a/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
+++ b/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
@@ -1,8 +1,6 @@
 ﻿using System.Text.Json;
 using Altinn.AccessManagement.UI.Core.ClientInterfaces;
-using Altinn.AccessManagement.UI.Core.Services.Interfaces;
 using Altinn.Platform.Register.Models;
-using Microsoft.AspNetCore.Http;
 
 namespace Altinn.AccessManagement.UI.Mocks.Mocks
 {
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -2449,9 +2458,9 @@ __metadata:
   linkType: hard
 
 "caniuse-lite@npm:^1.0.30001517":
-  version: 1.0.30001520
-  resolution: "caniuse-lite@npm:1.0.30001520"
-  checksum: 59991ad8f36cf282f81abbcc6074c3097c21914cdd54bd2b3f73ac9462f57fc74e90371cd22bcdff4d085d09da42a07dcea384cb81e4ac260496e1bd79e1fe7c
+  version: 1.0.30001521
```
Albertlarsen, line 2461: Revert changes in yarn.lock, there shouldn't be changes in this file when there isn't changes in package.json

allinox: That's not entirely true. Since we're using ^ (caret) on many of the package versions, we tell yarn to install that version or the latest minor or patch version, which is what has happened here.

Albertlarsen (approved):
🤌

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
```diff
@@ -9,6 +11,20 @@ namespace Altinn.AccessManagement.UI.Mocks.Mocks
...
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RegisterClientMock"/> class
+        /// </summary>
+        /// <param name="httpClient">http client</param>
+        /// <param name="httpContextAccessor">the handler for httpcontextaccessor service</param>
+        /// <param name="accessTokenProvider">the handler for access token generator</param>
+        public RegisterClientMock(
+            HttpClient httpClient,
+            IHttpContextAccessor httpContextAccessor,
```
Albertlarsen, line 22: Redundant line
```suggestion
           
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
```diff
@@ -9,6 +11,20 @@ namespace Altinn.AccessManagement.UI.Mocks.Mocks
...
+            IAccessTokenProvider accessTokenProvider)
```
Albertlarsen, line 23: redundant line
```suggestion
            )
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Mocks/Mocks/RegisterClientMock.cs
```diff
@@ -9,6 +11,20 @@ namespace Altinn.AccessManagement.UI.Mocks.Mocks
...
+        {
+        }
+        public RegisterClientMock() { }
```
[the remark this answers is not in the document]
allinox: Yeah, the Lookup test uses this parameterless constructor, (which was generated automatically before I added one with parameter. Now it has to be added directly in order to run the tests)

PR #569, opened 2023-09-29
make and use new component GroupElement

Albertlarsen:
## Description
- Make component GroupElements that takes react components as children and renders the elements besides eachother.
- I think the API of GroupElements is better than DualElements and NavigationButtons component as the component is more flexible.
- Delete DualElementsContainer and NavigationButtons
- Move all navigation buttons to the left
- Added "finished"-button to single rights ReceiptPage

## Related Issue(s)
- #542 

## Verification
- [ ] All tests run green

## Documentation


src/components/GroupElements/GroupElements.tsx
```diff
@@ -0,0 +1,35 @@
...
+ *  <Button>Previous</Button>
+ *  <Button>Next</Button>
+ * </GroupElements>
+ */
+
+export const GroupElements = ({ children }: GroupElementsProps) => {
+  return (
+    <ul className={classes.list}>
+      {children.map((child: React.ReactNode, index: number) => (
+        <li
```
allinox, line 26: Why display them as list items?

src/features/apiDelegation/offered/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -30,6 +30,7 @@ export const ConfirmationPage = () => {
   }, [loading]);
 
   const handleConfirm = () => {
+    console.log('Denne gjøres');
```
allinox, line 33: Did you mean to push this? 😅

src/localizations/en.json
```diff
@@ -32,7 +32,8 @@
     "read": "Read",
     "write": "Write",
     "sign": "Sign",
-    "access": "Access to whole service"
+    "access": "Access to whole service",
```
allinox, line 35:
```suggestion
    "access": "Access to the whole service",
```

allinox (approved):
Good job! 😊
I like the new groupElements component - I agree that it is a much more elegant solution than the previous two components

src/components/GroupElements/GroupElements.tsx
Albertlarsen: Design system does that for chip group and I think it is a common way to show semantically that two components are within the same group.

src/features/apiDelegation/offered/ConfirmationPage/ConfirmationPage.tsx
Albertlarsen: Hehe no

src/components/GroupElements/GroupElements.tsx
allinox: Ok, that makes sense 👍

PR #577, opened 2023-10-05
Update service resource model

allinox:
## Description

Updates the service resource model in the BFF and frontend and makes related changes to the mock data

## Related Issue(s)
- #559

## Verification
- [ ] All tests run green

## Documentation


backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Data/ExpectedResults/SingleRight/CreateDelegation/a3-app2.json
```diff
@@ -22,7 +22,7 @@
     {
       "resource": [
         {
-          "id": "urn:altinn:applicationid",
+          "id": "urn:altinn:app",
```
Albertlarsen, line 25: All services are now green on both ChooseServicePage and ChooseRightsPage. So there's probably some stuff that needs to be updated there as well.

Albertlarsen (changes requested):
All services are now green on both ChooseServicePage and ChooseRightsPage. So there's probably some stuff that needs to be updated there as well.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Data/ExpectedResults/SingleRight/CreateDelegation/a3-app2.json
allinox: Fixed! 😊

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Utils/AssertionUtil.cs
```diff
@@ -126,7 +126,7 @@ public static void AssertDelegationOutputEqual(DelegationOutput expected, Delega
         /// </summary>
         /// <param name="expected">An instance with the expected values.</param>
         /// <param name="actual">The instance to verify.</param>
-        public static void AssertResourceExternalEqual(ServiceResourceFE expected, ServiceResourceFE actual)
+        public static void AssertResourceEqual(ServiceResourceFE expected, ServiceResourceFE actual)
```
Albertlarsen, line 129: Don't understand why we should have unique names for all of the Assert methods, in Backend they just call them AssertEqual and let the parameters define which Asserts that's used. I think that's a simpler approach as you don't need to go into the AssertionUtil to check what the different Assertion methods are. I would have understood unique method name for a method that has exactly the same parameters and the name of that method being the same, but not in this case.

Albertlarsen (changes requested):

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Utils/AssertionUtil.cs
allinox: Honestly, this with different names for the assertion methods was a convention introduced by Dhana (and I expect they were lifted directly from backend) and I've just been maintaining it.
I don't have a very strong opinion about whether or not to keep the convention but if we're going to change it, I think it should be done in a separate task since it impacts a lot of methods

Albertlarsen: You can still call this method AssertEqual though

allinox: But that would mean breaking the convention that every other method in the file follows.
Changing the convention should be its own task imo so that we can have a clean change and actually finish the job.

(Also, the only thing done on this line here in this PR is to remove the 'External' part of the name, which never made sense to have here anyway but which was inherited from backend (where they do use external models) when this code was moved from the backend repo to here.)

Albertlarsen: This really takes more time discussing than doing it, made a PR that renames all the assertion methods.

https://github.com/Altinn/altinn-access-management-frontend/pull/580

Albertlarsen (changes requested):
Change Assertion method name

allinox:
@Albertlarsen 
The assertion method names are changed in a separate PR. It makes no sense to make the same changes here as well

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Utils/AssertionUtil.cs
allinox: Great!
Then that issue is handled separately

Albertlarsen (approved):

PR #580, opened 2023-10-10
Renamed Assertion method names

Albertlarsen:
## Description
Renames all assertion methods except AssertCollection to AssertEqual for easier test assertions.

## Verification
- [ ] All tests run green

## Documentation


backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/MaskinportenSchemaControllerTest.cs
```diff
@@ -53,7 +53,7 @@ public async Task GetReceivedMaskinportenSchemaDelegations_Valid_CoveredBy()
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
 
             List<MaskinportenSchemaDelegationFE> actualDelegations = JsonSerializer.Deserialize<List<MaskinportenSchemaDelegationFE>>(responseContent, options);
-            AssertionUtil.AssertCollections(expectedDelegations, actualDelegations, AssertionUtil.AssertMaskinportenSchemaDelegationFEEqual);
+            AssertionUtil.AssertCollections(expectedDelegations, actualDelegations, AssertionUtil.AssertEqual);
```
allinox, line 56: AssertCollection takes the name of the method to be used for assertion on that collection.
Since every method now has the same name, maybe AssertCollection should be updated so that we don't have to send in the same prop for every occation?

Albertlarsen: This is intended because when using that method `AssertEqual(expected, actual, AssertEqual)` I think it can confuse person with inception writing tests compared to `AssertCollection(expected, actual, AssertEqual)`. It can also makes sense given that compared to the other methods this method takes in collections and has three parameters.

allinox: I don't understand that first argument 😅 And the only reason why this takes three parameters is because it takes this assertMethodName as the third parameter which now is always the same name anyway.

That being said, there may be a use-case for recursive calling of AssertCollection, in which the name will have to be given.
How about a solution where the assertMethodName has AssertEqual as default value, so it doesn't have to be set unless something other than AssertEqual is used (which will be 98% of the time)?

Albertlarsen: Thought you meant something else, smart idea I'll try it 😊

allinox: Couldn't figure out how to do this after all since default values must be 'compile-time constants'
Shelving this idea and the thread

allinox (approved):

PR #627, opened 2023-11-22
New local list component as a wrapper for the Designsystem List

allinox:
## Description

This bug resulted from an update in the design system, which changed the List components styling and removed the option of bordered separation of list items.

This PR reintroduces this option by adding a wrapping component, that wraps the design system List in a component BorderedList, to provide the same styling as previously.

## Related Issue(s)
- #625 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/components/BorderedList/BorderedList.tsx
+++ b/src/components/BorderedList/BorderedList.tsx
@@ -1,0 +1,30 @@
+import type { ListProps } from '@digdir/design-system-react';
+import { List } from '@digdir/design-system-react';
+import cn from 'classnames';
+import * as React from 'react';
+
+import classes from './BorderedList.module.css';
+
+export type BorderedListProps = {
+  /**
+   * Style of the border separating the items
+   */
+  borderStyle?: 'solid' | 'dashed';
+} & ListProps;
+
+export const BorderedList = ({
+  borderStyle = 'dashed',
+  children,
+  className,
+  ...rest
+}: BorderedListProps) => {
+  return (
+    <List
+      className={cn(classes.borderedList, { [classes[borderStyle]]: borderStyle }, className)}
+      style={{ paddingLeft: 0 }}
+      {...rest}
+    >
+      {children}
+    </List>
+  );
+};
--- a/src/components/BorderedList/BorderedListItem/BorderedListItem.tsx
+++ b/src/components/BorderedList/BorderedListItem/BorderedListItem.tsx
@@ -1,0 +1,31 @@
+import type { ListItemProps } from '@digdir/design-system-react';
+import { List } from '@digdir/design-system-react';
+import cn from 'classnames';
+import * as React from 'react';
+
+import classes from './BorderedListItem.module.css';
+
+export type BorderedListItemProps = {
+  /**
+   * Style of the border separating the items
+   */
+  borderStyle?: 'solid' | 'dashed';
+} & ListItemProps;
+
+export const BorderedListItem = ({
+  borderStyle = 'dashed',
+  children,
+  ...rest
+}: BorderedListItemProps) => {
+  return (
+    <List.Item
+      className={cn(classes.borderedListItem, { [classes[borderStyle]]: borderStyle })}
+      style={{ listStyle: 'none', marginBottom: 0 }}
+      {...rest}
+    >
+      {children}
+    </List.Item>
+  );
+};
+
+BorderedListItem.displayName = 'BorderedList.Item';
--- a/src/components/BorderedList/BorderedListItem/BorderedListItem.tsx
+++ b/src/components/BorderedList/BorderedListItem/BorderedListItem.tsx
@@ -20,7 +20,7 @@ export const BorderedListItem = ({
   return (
     <List.Item
       className={cn(classes.borderedListItem, { [classes[borderStyle]]: borderStyle })}
-      style={{ listStyle: 'none', marginBottom: 0 }}
+      style={{ marginBottom: 0 }}
       {...rest}
     >
       {children}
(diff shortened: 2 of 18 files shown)


src/features/apiDelegation/offered/ReceiptPage/ReceiptPage.module.css
```diff
@@ -1,3 +1,3 @@
-.listContainer {
-  margin-bottom: 40px;
+.list {
+  margin-bottom: 40px !important;
```
Albertlarsen, line 2: !important is the absolutely last thing we should use to achieve the css we want as this in future will make it really hard for us to see which css-files that sets the styling. Did you explore other options?

src/components/BorderedList/BorderedListItem/BorderedListItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const BorderedListItem = ({
+  borderStyle = 'dashed',
+  children,
+  ...rest
+}: BorderedListItemProps) => {
+  return (
+    <List.Item
+      className={cn(classes.borderedListItem, { [classes[borderStyle]]: borderStyle })}
+      style={{ listStyle: 'none', marginBottom: 0 }}
```
Albertlarsen, line 23: I don't see why we should have classes and set style prop if we can do it all with one of those. Can't you set these styles in css classes as well?

src/components/BorderedList/BorderedList.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const BorderedList = ({
+  borderStyle = 'dashed',
+  children,
+  className,
+  ...rest
+}: BorderedListProps) => {
+  return (
+    <List
+      className={cn(classes.borderedList, { [classes[borderStyle]]: borderStyle }, className)}
+      style={{ paddingLeft: 0 }}
```
Albertlarsen, line 24: I don't see why we should have classes and set style prop if we can do it all with one of those. Can't you set these styles in css classes as well?

Albertlarsen (changes requested):

src/components/BorderedList/BorderedList.tsx
```diff
@@ -0,0 +1,30 @@
...
+import cn from 'classnames';
+import * as React from 'react';
+
+import classes from './BorderedList.module.css';
+
+export type BorderedListProps = {
+  /**
+   * Style of the border separating the items
+   */
+  borderStyle?: 'solid' | 'dashed';
```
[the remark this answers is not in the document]
allinox: Because if you don't set it, it defaults to dashed - I just figured it would be nice not to _have_ to set it to dashed all the time, when that's by far the one we use the most anyway

src/features/apiDelegation/offered/ReceiptPage/ReceiptPage.module.css
Albertlarsen: > This was the best way I could think of and !important isn't strictly bad, it exists for a reason,

Using !important is considered bad practise and should be the last resort after trying everything else. I will try to solve it another way.

Albertlarsen: You can just wrap the list in a div and add margin to the div or add margin-top to the elements below the list to achieve the same margin

allinox: Personally, I think the use of important in this case is a more elegant and clean solution than adding extra divs whose only purpose is adding margins.
It means far more clutter and I cannot see how the important will harm anything in this particular case

Albertlarsen: It has it's use case for example for Wordpress where it's really hard to style the web page in other ways, it just doesn't have a use case here where we _can_ solve it in other ways. And bad practise code isn't always removed from the programming languages because it's the way you use it, that defines if it's bad practise or not.

If you don't like adding extra divs you can just apply margin-top to the elements below the list to achieve the same result.

Albertlarsen: Also, taking potential problems vs code elegancy I think creating solutions that doesn't create potential problems outweighs elegancy by far.

I worked in a code base where it was used way too much important and that made it really hard to style the web site like you wanted because css priority didn't mean a thing there and you would just need to guess where the css containing the importants would be

Albertlarsen: Also when googling "css important" you will find almost everybody saying it should only be used as last resort. Even w3 has it stated in the article about the property 😅

image

src/components/BorderedList/BorderedList.tsx
allinox: I briefly touched upon this in the other thread, but the reason here is that we're battling the internal CSS of List, which adds a left padding we want to erase.

The only way to do this (that I know of) is by overpowering it in terms of priority, meaning either inline styling or with an !important tag, and while I've argued for the use of !important in the other thread, I don't think we should use it here as it is not the proverbial "end of the line". This is a reusable component, and it isn't unreasonable to assume there might be cases of conflicting css down the line.

src/features/apiDelegation/offered/ReceiptPage/ReceiptPage.module.css
allinox: That being said, I can see that we disagree on this point, and this hill is not one I'm gonna die on.
So I made the change to a wrapping-div anyway, which will be up in a minute

Albertlarsen (commented):
Great job 👍

Albertlarsen (approved):

Issue #619, opened 2023-11-15
Display only one action on A3 Apps in ChoosRightsPage


allinox:
### Description

Right now, all actions are displayed for all services. But with no handling of sub-resources or process steps in the Altinn Apps, this means that we can get duplicate actions belonging to different process steps or even unknown actions.

This is the current state of things:

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/43e90efb-2519-4bf8-950d-5af318c92098)

Until we can support multiple process steps and localized labels for all actions, we will handle Apps as a special case when displaying actions.
Instead of displaying all the actions individually, we will instead display a single action called "Access" ("Tilgang" på norsk) and hide all the underlying (true) actions.

This gives the following cases:

- User can delegate all actions on the app: Display "Access" as the only action. When delegation occurs, all the underlying accesses are delegated one by one.
- User can delegate _some_ actions on the app: Display "Access" as the only action and do not inform the user that there are some things they cannot delegate (the actionBar is green and with no alert). When delegation occurs, delegate all the underlying actions except the ones that the user does not have the rights to (as usual).
- User cannot delegate _any_ actions on the app: As with other resources, the actionBar should turn red in the ChooseService page and should not proceed to the ChooseRights page.

Albertlarsen:
Can we move this from review @allinox?

allinox:
Sorry, I forgot 😅

hannekristin:
@allinox Er det noe som skal sjekkes på denne saken? Er det en sak som er eldre enn #599 ?

allinox:
@hannekristin 
Denne er nok nyere enn #599 - det denne gjør er at alle apper displayer kun én rettighet som kan delegeres selv om de teknisk sett har flere.
Det ser slik ut:

![Image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/a8057753-5e41-46ee-8e79-8507a1df9fba)

hannekristin:
Testet ok

PR #660, opened 2023-12-12
Altinn apps now only displays one action: 'Access'

allinox:
## Description

Services of type AltinnApps only deisplays one single action, 'Access', on both rights page and receipt.
If some actions cannot be delegated or failed during delegation, these actions are not displayed in an alert and the user is not informed that all rights were not/will not be delegated unless no actions can be/were delegated.
During delegation, all rights that can be delegated will be delegated even though they were not specifically listed to the user.

This is a temporary solution that will last until the resource registry offers us a way to display sub-resources/tasks on apps.

## Related Issue(s)
- #619

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/dataObjects/dtos/singleRights/DelegationInputDto.tsx
+++ b/src/dataObjects/dtos/singleRights/DelegationInputDto.tsx
@@ -7,10 +7,12 @@ export interface DelegationInputDto {
 export class ServiceDto {
   serviceTitle: string;
   serviceOwner: string;
+  serviceType: string;
 
-  constructor(serviceTitle: string, serviceOwner: string) {
+  constructor(serviceTitle: string, serviceOwner: string, serviceType: string) {
     this.serviceTitle = serviceTitle;
     this.serviceOwner = serviceOwner;
+    this.serviceType = serviceType;
   }
 }
 
--- a/src/components/UserInfoBar/UserInfoBar.tsx
+++ b/src/components/UserInfoBar/UserInfoBar.tsx
@@ -17,8 +17,10 @@ export const UserInfoBar = () => {
   const dispatch = useAppDispatch();
 
   useEffect(() => {
-    if (userLoading || reporteeLoading) {
+    if (userLoading) {
       void dispatch(fetchUserInfo());
+    }
+    if (reporteeLoading) {
       void dispatch(fetchReportee());
     }
   }, []);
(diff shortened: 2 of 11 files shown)


src/dataObjects/dtos/singleRights/DelegationInputDto.tsx
```diff
@@ -7,10 +7,12 @@ export interface DelegationInputDto {
 export class ServiceDto {
   serviceTitle: string;
   serviceOwner: string;
+  serviceType: string;
 
-  constructor(serviceTitle: string, serviceOwner: string) {
+  constructor(serviceTitle: string, serviceOwner: string, serviceType: string) {
```
Albertlarsen, line 12: I started wondering if we could only show one action for Apps with a BFF solution instead. I want to massage the data in BFF more specifically for frontend in future so frontend can be more about just displaying the data it receives from BFF and do as little logic as possible. It made sense to do logic in frontend when we didn't have BFF, but now that we have. we need to make a transistion of how we do stuff.

The thing we would need to do here is making the service class actually do business logic and make apps just display action Access.

I just wanted to start a discussion about it because we have a lot of logic in frontend that we can start moving out to BFF. We can have a little discussion of pros and cons about doing it for this feature.

The biggest pro of this is that we don't get spread out logic in x numbers of components and Pages in frontend that it's hard to develop further in future because of all the dependencies.

Albertlarsen (changes requested):

src/dataObjects/dtos/singleRights/DelegationInputDto.tsx
allinox: I'm not a fan of this logic being spread out all over the place either, but this is a temporary solution and, unfortunately, I don't think we can easily make the BFF handle the logic in this case 😅

Because even if we're displaying only a single action, we still need to _remember_ the actual actions so that we can delegate them later. If the BFF is to do this, it needs user-individual memory, which it doesn't have.

Which is why I think it's better to keep this logic in frontend

Albertlarsen: Ahh that's true, however I wrote a draft user story that we can add tasks to when we come across some #667 
Just resolve this comment when read.

Albertlarsen (approved):

Issue #658, opened 2023-12-11
Single-rights | Failing resources are displayed in the wrong place


Albertlarsen:
### Description of the bug

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/977b075c-09f2-471d-afbf-2986b21142e6)

On the receipt page, the results of the delegations are displayed wrong - specifically, the failing actionBars are in the wrong place.

They should appear underneath the opened actionBar, so that all red actionBars are stacked together (underneath the text informing the user that there was one or more error) with the first actionBar defaulting as open.

What's interesting is that this bug doesn't appear when running locally with the mock. This task will thus be to figure out why the receipt malfunctions in AT and fix it.
The result should look like this:

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/b8725dc0-1d03-4668-abb0-45ec39c99c93)

### Steps To Reproduce

1. Go to AT23
2. Do single right delegation work flow and choose around 10 services
3. Observe that the failing resources are in the wrong place

Albertlarsen:
Can we move this from review @allinox?

PR #661, opened 2023-12-12
Failing resources displayed in the wrong order

allinox:
## Description

This bug was triggered when all failing delegations were returned as HTTP-errors, causing the error count to be 1 (as in the whole service) for every service. This revealed a flaw in the logic set up in the receipt page where two states were not updated correctly because of the limited scope inside of a forEach loop.
The loop relied on these states to decide which resource had the most 'failures' and thus would be displayed first (with default open). Since forEach binds the function passed to it, it limits the scope and will not get updated state values with each new iteration. This caused the loop to fire an update with every iteration which ended up marking the last entry in the list as the one to display first.

But since the list of delegations is then sorted separately (and every delegation failed with equal number) this stored index did not match the sorting of the delegations, causing the bug.

This is now fixed.

## Related Issue(s)
- #658 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/features/singleRight/delegate/ReceiptPage/ActionBarSection/ActionBarSection.tsx
+++ b/src/features/singleRight/delegate/ReceiptPage/ActionBarSection/ActionBarSection.tsx
@@ -2,7 +2,7 @@ import { useTranslation } from 'react-i18next';
 import { ErrorMessage, Ingress, Paragraph } from '@digdir/design-system-react';
 import * as React from 'react';
 import { ExclamationmarkTriangleIcon } from '@navikt/aksel-icons';
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 
 import { useAppSelector } from '@/rtk/app/hooks';
 import type { Right, ProcessedDelegation } from '@/rtk/features/singleRights/singleRightsSlice';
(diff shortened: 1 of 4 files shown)


src/features/singleRight/delegate/ReceiptPage/ActionBarSection/ActionBarSection.tsx
```diff
@@ -31,18 +31,24 @@ export const ActionBarSection = () => {
   });
 
   // logic for finding the delegation with the most failed Delegations
-  delegations.forEach((pd: ProcessedDelegation, index) => {
-    const failedDelegations = pd.bffResponseList?.filter(
-      (data: Right) => data.status !== BFFDelegatedStatus.Delegated,
-    );
+  useEffect(() => {
```
Albertlarsen, line 34: No related to your PR but it seems like the text key isn't localized anymore. Seems like something has gone wrong with some merge. Would be sweet if you could update the text key.

image

Albertlarsen (changes requested):
Good initiative. There's one text key that suddenly has changed in a merge or somthing I guess.

Albertlarsen (approved):
Good job!

PR #673, opened 2023-12-22
Feat/api delegation/read api id from url

Albertlarsen:
Important info:
A button click will add 2 of the same APIs in the chosen list because we use React.StrictMode in development environment. However if you comment out React.StrictMode you will be able to test how it actually will be like in prod, since React.StrictMode doesn't render application 2 times in prod.

## Description
Adds implementation of reading and writing api-ids to url. 

## Related Issue(s)
- #639 
- #641 
- #638 

## Verification
- [ ] All tests run green

## Documentation


src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -63,6 +61,39 @@ export const ChooseApiPage = () => {
...
+  const handleRemove = (api: DelegableApi) => {
+    removeApiFromParams(api);
+    dispatch(filter(filters));
+    dispatch(search(searchString));
+  };
+
+  const removeApiFromParams = (api: DelegableApi) => {
+    urlParams.delete(api.identifier);
+    setUrlParams(urlParams);
+    dispatch(softRemoveApi(api));
```
allinox, line 94: Very small thing but why is softRemoveApi in the removeApi_FromParams_ method?
I would expect it to be in the handleRemove method, after removeApiFromParams has finished.

allinox (approved):
Brilliant work! ✨
I've tried to be thorough with my testing, and as far as I can see, it works really well 😄

Issue #622, opened 2023-11-16
Use new Search component from the Design system


allinox:
### Description

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/8fa84f87-8091-42be-b426-148293d9c25d)

The Design system has created a new Search component, with better features and accessibility than the one we currently use (from the Altinn design system). We should thus switch to using this one wherever we use search

One thing we need to determine is if we should continue using a "hot reload" strategy for our search or use the new "submit" feature, that the new component facilitates.

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/6392de9d-97f3-4e04-9ae3-fe4d7048b9cb)

Different strategies may be used for different places.

## Conclusion

We will use the search field without button as we do today.
This is because fast feedback is nice for the user.

To limit the number of search calls, however, we will implement a debounce function to be used on the search field.
This way we can delay reaction to change in the search query by a few milliseconds.

### Places to change:

- API delegation:
  - Org search
  - API search
- Single rights:
  - Service search

hannekristin:
Testet ok

PR #680, opened 2024-01-03
Feat/general/622 use new search component

allinox:
## Description
All search fields have been switched to the new designsystem search component and a debounce function has been added to save computational power. There is currently a bug in the designsystem component though, where the input field is stuck at medium size and only the icon changes. Because of this, I've set the component to 'medium' for now, even if this is larger than the previous input field.

Also discovered a couple of bugs in the filter-component that came as a result of style changes in the design system. These have now been fixed.

## Related Issue(s)
- #622

## Verification
- [ ] All tests run green

## Documentation


src/resources/utils/debounce.ts
```diff
@@ -0,0 +1,12 @@
+// Delays calling of func by timeout ms
+export function debounce(func: (...args: any[]) => void, timeout = 300) {
```
Albertlarsen, line 2: Why is this a function and not a const?

src/resources/utils/debounce.ts
```diff
@@ -0,0 +1,12 @@
+// Delays calling of func by timeout ms
```
Albertlarsen, line 1: Reusable functions and components should have jsDoc

Albertlarsen (changes requested):
The Search component looks super nice, good work 😊

src/components/Filter/OptionDisplay.module.css
```diff
@@ -11,20 +11,20 @@
...
   width: 100%;
   border: none;
-  border-bottom: 1px solid var(--fds-semantic-border-neutral-divider);
+  border-bottom: 1px solid var(--fds-semantic-border-divider-default);
   background-color: var(--fds-semantic-background-default);
   cursor: pointer;
   font: inherit;
   text-align: left;
-  padding: 0;
+  padding: 4px 0 4px 0;
```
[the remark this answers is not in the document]
allinox: You're right! I'll do that 😊

src/resources/utils/debounce.ts
```diff
@@ -0,0 +1,12 @@
...
+export function debounce(func: (...args: any[]) => void, timeout = 300) {
+  console.log('ready to bounce');
```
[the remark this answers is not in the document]
allinox: Oops 😅

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -148,6 +148,10 @@ export const ChooseApiPage = () => {
     );
   });
 
+  const debouncedSearch = debounce((searchString: string) => {
```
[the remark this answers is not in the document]
allinox: The reason it's done this way is because the function that is to be debounced needs to be bounded in order to fire with the right parameter - it needs the context surrounding it, and the easiest way to provide that is by binding it with an arrow function. It wouldn't have to be this way if the debounce-function was implemented right here in this file, but since I've made it into a reusable function, the context for the handleSearch-function needs to be provided so that it can be fired from the different file/context that the debounce-function's code resides in.

src/resources/utils/debounce.ts
allinox: Because the declaration of const is used with arrow functions, and this is not an arrow function

allinox: I refrained from using an arrow function here since binding is so important and I didn't want to cross any wires

src/features/singleRight/components/SearchSection/SearchSection.tsx
```diff
@@ -227,19 +235,30 @@ export const SearchSection = ({ onAdd, onUndo }: SearchSectionParams) => {
...
+    setSearchString(searchString);
+    setCurrentPage(1);
+  }, 300);
+
   return (
     <div className={classes.searchSection}>
       <div className={classes.searchInputs}>
         <div className={classes.searchField}>
-          <SearchField
+          <Search
```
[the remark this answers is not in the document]
allinox: Good catch!
Jeg kan høre med de andre i designsystemet 😊

src/resources/utils/debounce.ts
Albertlarsen: Isn't it possible to bind with arrow function?

Albertlarsen: Doesn't matter that much since this is a standalone reusable function, I'll resolve it

Albertlarsen (approved):
Nicely done 😊

PR #683, opened 2024-01-04
Feat/api delegation/642 skeleton for choose api page

Albertlarsen:
## Description
Adds Skeleton for ChooseApiPage in relation to deeplinking and delegationCheck.

## Related Issue(s)
- #642 

## Verification
- [ ] All tests run green

## Documentation


src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -75,6 +70,7 @@ export const ChooseApiPage = () => {
         }
       });
     }
+    setDelegationCheckLoading(false);
```
allinox, line 73: This needs to be set after we've received the results from the delegationChecks. Right now, the skeletons disappear before the changes happen on screen, which means that we see the actionBars go from one side to the other.

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -167,17 +163,18 @@ export const ChooseApiPage = () => {
...
-        removeCallback={chosenOrgs.length > 1 ? () => dispatch(softRemoveOrg(org)) : null}
-        leftText={org.orgName}
-        middleText={!isSm ? t('api_delegation.org_nr') + ' ' + org.orgNr : undefined}
-      ></CompactDeletableListItem>
+      <>
+        <Skeleton.Rectangle height={skeletonHeight}></Skeleton.Rectangle>
+        <Skeleton.Rectangle height={skeletonHeight}></Skeleton.Rectangle>
+        <Skeleton.Rectangle height={skeletonHeight}></Skeleton.Rectangle>
+        <Skeleton.Rectangle height={skeletonHeight}></Skeleton.Rectangle>
+        <Skeleton.Rectangle height={skeletonHeight}></Skeleton.Rectangle>
```
allinox, line 174: I suggest removing one of these rectangles, since 5 large blocks on both sides feels a bit heavy

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
allinox: I guess that means we have to either put this state into the redux state or make delegationCheck into a rtkQuery api

allinox (changes requested):

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
Albertlarsen: Ahh yeah you're right. Best thing would be if we could await the last call on delegationCheck or something. Thinking about it, that's something we maybe could have done if delegationCheck was a mutation in RTK Query instead of being a slice? I Haven't analyzed it yet, but it may work. If that's something you want to try out when I'm on vacation you can just take over this whole task as you please 😊

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
Albertlarsen: I don't discuss or apply code using feelings as guideline and if we were to discuss feelings all the time but if I were to use feelings as an argument I could have just said that I felt 5 large blocks was just about right as an counterargument and the discussion would stall.

However the reason why there are 5 large blocks there is because when the view is done loading it is displayed 5 ActionBars there. However I can agree on having these 5 rectangles on both sides for "delegerbare api" and "valgte API" is unneccessary and we could reduce to having them just on the delegerbare api-side or just have 5 action bars spanning over the whole container.

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -187,29 +184,16 @@ export const ChooseApiPage = () => {
...
-          ) : (
-            <div>
-              <h3>{t('api_delegation.chosen_orgs')}:</h3>
-              <BorderedList borderStyle={'dashed'}>{chosenDelegableOrgs}</BorderedList>
-            </div>
-          )}
           <h3 className={classes.chooseApiSecondHeader}>
             {t('api_delegation.new_api_content_text2')}
           </h3>
           {isSm && chosenApis.length > 0 && (
```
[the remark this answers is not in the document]
Albertlarsen: The Skeletons for delegerbare api are shown on smaller screens but not valgte api skeletons and I think that is the correct approach as showing skeletons for both would take up so much space on the page.

image

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
allinox: I'm not talking about feeling as in emotions here, I'm talking about the impression it has on the user.
10 large blocks of grey, with minimal padding between them, is very heavy visually, pulling focus and taking up space like a brick wall in the middle of the page - to the user, it will _feel heavy_.
This is especially true when most of this will be replaced with blankness a second later to it can end up feeling like a hit-and-run meeting with a truck.

Part of this can be helped by removed by using fewer rectangles, but it would be even to, as you say, removing the skeletons on the right (I just didn't think I'd get you to do that) 😅
Since there sometimes can be some pre-chosen actionBars, we could also consider having just one or two on the right side. Idk, I'd have to see the different options to see what works the best.

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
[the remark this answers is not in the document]
allinox: But the thing is, there _are_ skeletons added for the chosen api's, added just a couple of lines down.
Which means that once the first response of the delegationCheck comes back, these skeletons will appear and be on screen until the rest of the responses comes back.
So then we will have some skeletons appear, then some other skeletons will appear, and then all of them will disappear, which will look strange.

Now, considering the other discussion about whether to display skeletons for the chosen APIs at all, this might not be an issue. In that case, we will simply have to make sure nothing is displayed in this section until everything is displayed.

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
allinox: Yeah, since this will soon go into production, I think I'll sneak into this branch and try to implement delegationCheck with rtkQuery and mutations 😊

Issue #688, opened 2024-01-12
Frontend crashes in all AT environments


allinox:
### Description of the bug

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/d9d99497-da27-456d-9fe8-566d9d9fe1e8)

This only relates to the choose-service page in single-rights

### Steps To Reproduce

1. Go to https://am.ui.at22.altinn.cloud/accessmanagement/ui/delegate-single-rights/choose-service
2. Observe the BOD (Bluescreen Of Death)

allinox:
Update: After fixing the error page, the failing page looks like this:
![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/22affd3e-3c2f-432d-a891-840b2e7dbd35)

The error seems to be related to rtk query but I am still not able to recreate it locally.

allinox:
@anishastphen @acn-Sneha 
It's working again 🎉

PR #691, opened 2024-01-15
Robustifying query subscriptions

allinox:
## Description
The error seems to be related to rtk query, which seems to unsubscribe from its subcriptions for some unknown reasons, causing the error `N.current.isRequestSubscribed is not a function` to trigger from inside rtk. The fact that I am not able to recreate this error locally, makes debugging difficult, but in a blind attempt at fixing the error in AT, I have cleaned up some code, added a listener, and exported the middleware from the api in the hopes that this will make the subscriptions more robust

## Related Issue(s)
- #688 

## Verification
- [ ] All tests run green

## Documentation


jonkjetiloye (approved):
LGTM 👍

PR #697, opened 2024-01-19
Finalize Dyplenke with rtkQuery for delegationCheck and skeleton on loading

allinox:
## Description
This PR finilizes the dyplenke userstory by using skeleton on loading to hide the shuffling of the actionBars while the delegationCheck is happening. In doing so however, large rewrites of logic had to be performed, changing delegationCheck from being performed and stored in Redux state, to using rtkQuery. In doing so, the necessary logic and the fetched can be contained only to the area where it is needed, in the actionBar component representing the API's, now renamed into ApiActionBars.

## Related Issue(s)
- #642 

## Verification
- [ ] All tests run green

## Documentation


src/features/apiDelegation/components/ApiActionBar/ApiActionBar.tsx
```diff
@@ -0,0 +1,222 @@
...
+}: ApiActionBarProps) => {
+  const [open, setOpen] = useState(false);
+  const [actionBarColor, setActionBarColor] = useState<'success' | 'danger' | 'neutral'>(
+    variant === 'remove' ? 'success' : 'neutral',
+  );
+  const { t } = useTranslation('common');
+  const resourceRef: ResourceReference = { resource: api.authorizationReference };
+  const partyId = getCookie('AltinnPartyId');
+
+  const [checkCanDelegate, { data: accessResult, error, isLoading: isUpdating }] =
```
Albertlarsen, line 46: Like delegationCheck or canDelegateCheck better than name checkCanDelegate

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -46,6 +51,11 @@ export const ChooseApiPage = () => {
   const fetchData = async () => await dispatch(fetchDelegableApis());
   const navigate = useNavigate();
   const [urlParams, setUrlParams] = useSearchParams();
+  const [showSkeleton, setShowSkeleton] = useState(true);
+
+  const partyId = getCookie('AltinnPartyId');
+
+  const [checkCanDelegate] = useDelegationCheckMutation();
```
Albertlarsen, line 58: Like delegationCheck or canDelegateCheck better than name checkCanDelegate

src/features/apiDelegation/components/ApiActionBar/ApiActionBar.tsx
```diff
@@ -0,0 +1,222 @@
...
+    useDelegationCheckMutation();
+
+  const isNotDelegable = error || (accessResult && accessResult.status === 'NotDelegable');
+
+  useLayoutEffect(() => {
```
Albertlarsen, line 51: Nice detail

src/features/apiDelegation/components/ApiActionBar/ApiActionBar.tsx
Albertlarsen, line 46: imo isLoading is a clearer phrasing than isUpdating so I don't know why you converted it do isUpdating

Albertlarsen (changes requested):
Really nice work Alexandra 😊

src/features/apiDelegation/offered/ChooseApiPage/ChooseApiPage.tsx
```diff
@@ -123,44 +150,62 @@ export const ChooseApiPage = () => {
...
       );
     }
+
+    const prechosenApis = Array.from(urlParams.keys());
+
     return presentedApis.map((api: DelegableApi) => {
+      const initWithDelegationCheck = prechosenApis.includes(api.identifier);
       return (
-        <DelegationActionBar
+        <ApiActionBar
```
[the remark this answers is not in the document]
allinox: Shoot, I didn't think of that...

Yeah,, the ActionBars won't be able to store that themselves unfortunately, since they're basically killed and reborn every time the search changes.

The only way of reminding them of their past lives is to store who've been previously selected in the parent, so a local state in ChooseApiPage that stores a list of the chosen but not delegable apis. We could technically do that and use the `initWithDelegationCheck` to trigger them refetcing the mutation from cache, but that would mean adding yet another state just for this and... idk.. I'm not sure if I think it's worth it honestly 😅

What do you think?

src/features/apiDelegation/components/ApiActionBar/ApiActionBar.tsx
allinox: It was mostly because queries have isLoading and isFetching which represent the initial loading and the refetching of data respectively, while mutations only have isLoading representing both. rtkQuery's documentation suggested renaming it to isUpdating locally to make that more clear and I figured I'd do the same, but I guess in our case that distinction doesn't really matter so isLoading is better.
I'll change it 👍

Issue #733, opened 2024-02-15
Cypress tests are failing due to change in language takes longer time to get reflected


acn-Sneha:
Our cypress tests has pre-requisite : The choosen language should always be norwegian before it starts to run all tests but I see changing language takes 20 minutes to get reflected in altinn 3 which might be problematic for us to run cypress tests

acn-Sneha:
@Albertlarsen raised a bug as we discussed.

Albertlarsen:
@acn-Sneha Final fix of this was just made and for me language change seems to work instantly now. Can you verify that it works as expected?

PR #735, opened 2024-02-16
Read language cookie from altinnPersistentContext

Albertlarsen:
## Description
Adds code that tries to read language from AltinnPersistence cookie first.

## Related Issue(s)
- #733

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/Controllers/HomeController.cs
+++ b/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/Controllers/HomeController.cs
@@ -94,13 +94,39 @@ namespace Altinn.AccessManagement.UI.Controllers
             int userId = AuthenticationHelper.GetUserId(_httpContextAccessor.HttpContext);
             UserProfile user = await _profileService.GetUserProfile(userId);
             AntiforgeryTokenSet tokens = _antiforgery.GetAndStoreTokens(HttpContext);
-            string languageCode = ProfileHelper.GetStandardLanguageCodeForUser(user);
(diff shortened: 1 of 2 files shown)


allinox (commented):
Question: In the case where cashing is an issue, and the cached profile language differs from the cookie, won't we still run into problems when we internally fetch the language from profile?

I'm thinking about the method `GetLanguageCodeForUser` in profileHelper, which we use in several of our BFF controllers to get the active language.
Maybe this method should be updated as well or it's use replaced with something else? (Like a helper method for fetching language that checks the cookie first and profile second?)

allinox:
Oh no! 
I was 2 minutes too late with my comment! 😅

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/Controllers/HomeController.cs
```diff
@@ -94,14 +94,40 @@ private async Task SetLanguageCookie()
...
                 // Make this cookie readable by Javascript.
                 HttpOnly = false,
             });
         }
+        
+        private string GetAltinnPersistenceCookieValue()
+        {
+            var cookieValue = HttpContext.Request.Cookies["altinnPersistentContext"];
+
+            if (cookieValue.Contains("UL1033"))
```
allinox, line 114: Also, I think it should be `UL=1033`
image

Albertlarsen:
@allinox  Ahh that's true... Problem is that HomeController is static and I got error when trying to read cookie from a static class. I'll make something new and try to fix it

Issue #740, opened 2024-02-19
Recipient name is lost during delegation process


allinox:
### Description of the bug

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/a083481a-1099-4741-b3af-89b80798fd4a)

The recipient name is fetched by looking up the uuid gotten from the url. This fetching is cached in rtk query and then fetched from here for every page throughout the delegation process.
This seems to work well when the recipient is an org but when the recipient is a person, however, the name is lost for some unknown reason.

Figure out why this is and fix it.

Note: This seems to not be the case in local running

### Steps To Reproduce

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/5f81ffc8-6751-4136-a6a1-52a0f28d3fa7)
->
![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/ec9f83b1-5399-4bfc-b645-3b4d48aa321d)

PR #748, opened 2024-02-22
Fixed dropping recipient name

allinox:
## Description

This PR introduces skipping unnessesary calls to lookup souch as calls with invalid uuid (before uuid has been fetched from url) and calls whose data is not needed (lookup of person when recipient is a praty and lookup of party when recipient is a person).

It also fixes the main bug in #740 by triggering data logic in the fetching hook straight away (instead of on an update) which is nessesary in the cases where data is fetch from cache and thus happens immediately.
(Plus it adds a change to styling on receipt page for the case where all delegations are successfull and no extra padding is needed over the success message)

## Related Issue(s)
- #740 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/resources/hooks/useFetchNameFromUUID.tsx
+++ b/src/resources/hooks/useFetchNameFromUUID.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import {
@@ -6,8 +6,6 @@ import {
   useGetPartyByUUIDQuery,
   useGetUserByUUIDQuery,
 } from '@/rtk/features/lookup/lookupApi';
-
-import { useUpdate } from './useUpdate';
 
 // Used for fetching recipient name from userUUID or partyUUID
 export const useFetchNameFromUUID = (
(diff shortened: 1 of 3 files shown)


src/resources/hooks/useFetchNameFromUUID.tsx
```diff
@@ -18,20 +16,33 @@ export const useFetchNameFromUUID = (
 
   const [recipientName, setRecipientName] = useState<string | undefined>(undefined);
   const [error, setError] = useState(false);
+  const [skipUser, setSkipUser] = useState(userUUID === null);
+  const [skipParty, setSkipParty] = useState(partyUUID === null);
```
Albertlarsen, line 20: Instead of adding another state that needs to be maintained can't you use the userUUUID to see if you should skipParty?

Albertlarsen (changes requested):
Good job with identifying and solving the problem ☺️

src/resources/hooks/useFetchNameFromUUID.tsx
allinox: You're right!
I couldn't get it to work without states in the beguinning, but that must have been because of something else because now it suddenly works.
I'll push the change 😊

Albertlarsen (approved):

Issue #826, opened 2024-03-26
Clear Access Cache for recipient after new single right access delegation(s)


jonkjetiloye:
### Description

Access Management backend has a new internal API endpoint for invalidating access cache after a new delegation has been performed.

Endpoint URL: 
`PUT /accessmanagement/api/v1/internal/{{party}}/accesscache/clear`

Body examples:

```
{
  "type": "urn:altinn:person:uuid",
  "value": "00000000-0000-0000-0000-000000000000"
}
```

```
{
  "type": "urn:altinn:organization:uuid",
  "value": "00000000-0000-0000-0000-000000000000"
}
```

```
{
  "type": "urn:altinn:person:uuid",
  "value": "00000000-0000-0000-0000-000000000000"
}
```

Access management frontend must be updated so that after a all resource delegations to a recipient has completed the access cache clear API is triggered, which hopefully should trigger the necessary cache updates, so that the delegator is able to immediately see the new changes when navigating back to Altinn 2 profile page.

### Tasks

- [x] AccessManagement Client operation
- [x] BFF endpoint
  - [x] Mock endpoint
- [ ] Frontend logic for calling API once all delegation requests have completed

### Acceptance Criterias

As a main administrator/tilgangsstyrer for a party or myself
I should be authorized to clear access cache for a given delegation recipient identified by their uuid
once cleared the recipents reportees/authorized-parties and rights for the delegating party should be updated with the latest changes

PR #842, opened 2024-04-05
Clear Access Cache for recipient after new single right access delegation(s)

allinox:
## Description

This PR adds
- New endpoint in BBF for calling the new cache clearing endpoint in backend
- Calling of this BFF endpoint in frontend after all delegations have been performed
- Added new model BaseAttribute which will eventually replace IdValuePair as the main model for defining single attributes

## Related Issue(s)
- #826 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/ISingleRightService.cs
+++ b/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/ISingleRightService.cs
@@ -30,5 +30,17 @@ namespace Altinn.AccessManagement.UI.Core.Services.Interfaces
         /// </param>
         /// <returns></returns>
         Task<HttpResponseMessage> CreateDelegation(string party, DelegationInput delegation);
+
+        /// <summary>
+        ///     Clears cashed acesses of a user
+        /// </summary>
+        /// <param name="party">
+        ///     The party from which the rights have been given (delegator)
+        /// </param>
+        /// <param name="user">
+        ///     The uuid identifier of the user (delegation recipient) to clear access cashe on
+        /// </param>
+        /// <returns></returns>
+        Task<HttpResponseMessage> ClearAccessCasheOnUser(string party, BaseAttribute user);
     }
 }
--- a/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/SingleRightControllerTest.cs
+++ b/backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/SingleRightControllerTest.cs
@@ -439,6 +439,43 @@ namespace Altinn.AccessManagement.UI.Tests.Controllers
             Assert.Equal(HttpStatusCode.BadRequest, httpResponse.StatusCode);
         }
 
+        /// <summary>
+        ///     Test case: ClearAccessCashe accepts correct input and returns ok
+        ///     Expected: ClearAccessCashe returns OK
+        /// </summary>
+        [Fact]
+        public async Task ClearAccessCashe_returnsOk()
+        {
+            // Arrange
+            string partyId = "999 999 999";
+            string userUUID = "5c0656db-cf51-43a4-bd64-6a91c8caacfb";
+
+            List<IdValuePair> resource = new List<IdValuePair>
+            {
+                new IdValuePair
+                {
+                    Id = "urn:altinn:resource",
+                    Value = "Nonexistent",
+
+                },
+            };
+
+            BaseAttribute user = new BaseAttribute
+            {
+                    Type = "urn:altinn:person:uuid",
+                    Value = userUUID,
+            };
+
+            string jsonDto = JsonSerializer.Serialize(user);
+            HttpContent content = new StringContent(jsonDto, Encoding.UTF8, "application/json");
+
+            // Act
+            HttpResponseMessage httpResponse = await _client.PutAsync($"accessmanagement/api/v1/singleright/{partyId}/accesscache/clear", content);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, httpResponse.StatusCode);
+        }
+
         private int CountMatches(List<DelegationResponseData> actualResponses, string expectedResponseFileName)
         {
             string path = Path.Combine(mockFolder, "Data", "SingleRight", "DelegationAccessCheckResponse", expectedResponseFileName);
--- a/src/dataObjects/dtos/BaseAttribute.tsx
+++ b/src/dataObjects/dtos/BaseAttribute.tsx
@@ -1,0 +1,10 @@
+// This model describes an attribute, like a resource, a user, a party or an action. The attribute is defined by its type (usually a urn) and it's value.
+export class BaseAttribute {
+  type: string;
+  value: string;
+
+  constructor(type: string, value: string) {
+    this.type = type;
+    this.value = value;
+  }
+}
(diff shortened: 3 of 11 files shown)


src/dataObjects/dtos/BaseAttribute.tsx
allinox, line 1: @Albertlarsen 
If you have any suggestions on a better name for this model, I would very much like to hear it.
This model will eventually replace the IdValuePair model we use today and so far, I just gave it the same name as in backend. I'm not exactly overjoyed by this name, though, I just couldn't come up with a better alternative

src/features/singleRight/delegate/ChooseRightsPage/ChooseRightsPage.tsx
```diff
@@ -66,6 +69,7 @@ export const ChooseRightsPage = () => {
   const processedDelegationsRatio = (): number =>
     Math.round((processedDelegations.length / delegationCount) * 100);
   const buttonRef = useRef<HTMLButtonElement | null>(null);
+  const [clearAccessCashe] = useClearAccessCasheMutation();
```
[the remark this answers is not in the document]
Albertlarsen: I don't understand why this is a mutation, we don't use any data from that call? So a simple axios call would do. It's not critical for release but anyways

src/rtk/features/singleRights/singleRightsApi.ts
```diff
@@ -64,9 +62,22 @@ export const singleRightsApi = createApi({
     getResourceOwners: builder.query<ResourceOwner[], void>({
       query: () => 'resources/resourceowners',
     }),
+    clearAccessCashe: builder.mutation<void, { party: string; user: BaseAttribute }>({
```
Albertlarsen, line 65: I don't understand why this is a mutation, we don't use any data from that call? So a simple axios call would do. It's not critical for release but anyways

src/dataObjects/dtos/BaseAttribute.tsx
jonkjetiloye: This is also probably not the final model to rule them all for attributes in type, value format, as Alexander is working on a whole new library for better parsing and handling of urn's. Which is whats we will probably all end up using in the end.

Albertlarsen (commented):
Just some small feedback, if someone else approves it's okay just merge and we can fix the feedback comments later :)

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/ISingleRightService.cs
```diff
@@ -30,5 +30,17 @@ public interface ISingleRightService
...
+        ///     Clears cashed acesses of a user
+        /// </summary>
+        /// <param name="party">
+        ///     The party from which the rights have been given (delegator)
+        /// </param>
+        /// <param name="user">
+        ///     The uuid identifier of the user (delegation recipient) to clear access cashe on
+        /// </param>
+        /// <returns></returns>
+        Task<HttpResponseMessage> ClearAccessCasheOnUser(string party, BaseAttribute user);
```
jonkjetiloye, line 44: My only minor feedback comments (across implementation):
- Typo Cashe -> Cache
- BaseAttribute variable name 'user' would be more covering to use recipient, as this is either a user or a org party

jonkjetiloye (approved):

src/features/singleRight/delegate/ChooseRightsPage/ChooseRightsPage.tsx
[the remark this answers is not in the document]
allinox: It's a mutation because it's a PUT call (and it technically does change something on backend, namely the cache)
I put it in the SingleRightsApi as an rtkQuery because it's related to the rest of the SingleRights feature and part of the BFF's API's so this way it adheres to our already established pattern.
Not to mention that it's honestly just as easy to add it to the rtkQuery as it would have been to use an axios call (with the added benefit of abstracting the actual network call away from this file) 😊

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/SingleRightControllerTest.cs
```diff
@@ -439,6 +439,43 @@ public async Task CreateDelegation_StandardResource_invalid()
...
+        [Fact]
+        public async Task ClearAccessCashe_returnsOk()
+        {
+            // Arrange
+            string partyId = "999 999 999";
+            string userUUID = "5c0656db-cf51-43a4-bd64-6a91c8caacfb";
+
+            List<IdValuePair> resource = new List<IdValuePair>
+            {
+                new IdValuePair
```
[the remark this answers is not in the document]
allinox: Oops, this resource shouldn't be here. I forgot to remove it 😅
Good catch!

src/rtk/features/singleRights/singleRightsApi.ts
allinox: It's a mutation because it's a PUT call (and it technically does change something on backend, namely the cache)
I put it in the SingleRightsApi as an rtkQuery because it's related to the rest of the SingleRights feature and part of the BFF's API's so this way it adheres to our already established pattern.
Not to mention that it's about as easy to add it to the rtkQuery as it would have been to use an axios call (with the added benefit of abstracting the actual network call away from this file) 😊

src/dataObjects/dtos/BaseAttribute.tsx
allinox: In that case, I'll just leave it as it is for now

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/ISingleRightService.cs
allinox: Fixed! 😊

src/rtk/features/singleRights/singleRightsApi.ts
Albertlarsen: There's not an established pattern for this, as this is the first time we're doing cache invalidation in the frontend or BFF. Yes it is a HTTP request but different types of HTTP requests should have their own pattern.

The difference between this and the other HTTP requests we have in frontend, is that this request doesn't receive any object in return and hence I don't think it needs to be a whole mutation because if this had been written as axios-call instead it would have made it clear that this request doesn't receive any data related from the request (except of HTTPResponse, but every HTTP Request has HTTPResponse). If all our future HTTP requests that doesn't receive object had been axios calls it would have made it clear in code base that when we do axios calls we don't expect any data. Also axios is more lightweight than Redux (although when receiving object, redux is better)

And that is a pattern.

Albertlarsen: If this on the other hand was a Put request receving an object from the request then yeah I agree it would make sense this being a mutation

allinox: I don't understand...
Why create an entirely different standard for something that is already covered by an existing one in our setup?

When I'm talking about standard and pattern here, I mean how we do HTTP calls (GET, POST, PUT, DELETE) and our current way of doing that is through rtkQuery (or through rtkSlices, but we've talked about moving a lot of that over to rtkQuery anyway). It's organized, it's reusible, and adding something to an existing Api-file is easy as sugar. (And this specific endpoint is part of the Single-Rights API anyway, so it fits right into that same rtk api setup)

I just don't see how any of this would be affected by whether or not the call returns a body?

Albertlarsen:
For å gjøre det klart hva jeg mener om det her så er det at jeg synes det er et steg i feil retning å skulle løse Altinn 2-problemer i Altinn 3. Med dette så er den elven åpnet og da får vi se hvor mye som renner inn.

Issue #856, opened 2024-04-15
Can still delegate even when "Access" is deselected on A3 services


allinox:
### Description of the bug

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/1f9d363c-9ce8-4a3c-8f16-db59aa56ebda)

Normally, the fact that all chips are deselected would disable the button to trigger delegations. However, in the case of A3 apps, this chip is currently not directly linked to the actions for delegation, but is instead a catch all to enable delegation of all possible actions.

The error here is that the button for delegation should be disabled when this chip is deselected and there are no other services with selected chips. It is likely fixed by setting the selected attribute on all actions on the service to false/true on the service whenever this chip is deselected/selected

### Steps To Reproduce

1. Go to the single rights page
2. Choose an A3 app
3. Deselect "Tilgang" chip
4. Observe that you can still press "Fullfør delegering"

Albertlarsen:
Proposed fix: Legger til samle-chipen "tilgang" i den faktiske lista over actions som er koblet til koden for å disable "Fullfør delegering"-knappen.

Albertlarsen:
Refined proposed fix: Legg til rightskey-felt i rightKeyen som vi toggler på i stedet

PR #862, opened 2024-04-17
Bugfix/856 can still delegate app action access

Albertlarsen:
## Description
Fixed:
- Filtered out NotDelegable actions in apps in redux slice
- Added code for Apps when there's not any actions with errorCodes that it sets the error code and message to MissingRoleAccess on ResourceActionBar.
- Added test resource that's the same as in TT for apps
- Introduced jsonDoc comments on intern const which was very nice because now you can hover over the function to see what it does
- moved const toggleAllDelegableRights to SearchSection instead which is used for Apps to toggleAllDelegableRights for an App.

Ask me of any clearifications why I did what I did 😊

## Related Issue(s)
- #856 

## Verification
- [ ] All tests run green

## Documentation


src/resources/utils/debounce.ts
```diff
@@ -7,7 +7,7 @@
  * @returns {Function} A new function that delays the invocation of the input function.
  *
  * @example
- * debounce(() => console.log('Hello, world!'), 500);
```
Albertlarsen, line 10: had to change it for when I search for console.log it shows up and I think I have a console log in my code :p

Issue #861, opened 2024-04-17
Pagination goes to page 1 when filter is opened and closed


Albertlarsen:
### Description of the bug

Pagination goes to page 1 when filter is opened and closed

<img width="1070" alt="image" src="https://github.com/Altinn/altinn-access-management-frontend/assets/31471142/aa37093e-2250-4032-a28e-66a730293132">

### Steps To Reproduce

1. Go to another page in the pagination
2. Open filter
3. Close Filter

PR #884, opened 2024-04-24
Bugfix/875 rar oppforsel i filter i api delegering

Albertlarsen:
## Related Issue(s)
- #875 
- #861 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/components/Filter/OptionDisplay.tsx
+++ b/src/components/Filter/OptionDisplay.tsx
@@ -84,12 +84,12 @@ export const OptionDisplay = ({
     setSortedOptions(optionSearch(options, searchString));
   };
 
-  const checkboxes = sortedOptions.map((option) => {
+  const checkboxes = sortedOptions.map((option, index) => {
     const isSelected = selectedValues?.includes(option.value);
     return (
       <button
         className={classes.option}
-        key={option.label}
+        key={`filterOption_${index}_${option.value}`}
         onClick={() => {
           handleSelection(option.value);
         }}
--- a/src/components/Filter/OptionDisplay.tsx
+++ b/src/components/Filter/OptionDisplay.tsx
@@ -89,7 +89,7 @@ export const OptionDisplay = ({
     return (
       <button
         className={classes.option}
-        key={option.value}
+        key={option.label}
         onClick={() => {
           handleSelection(option.value);
         }}
--- a/src/features/singleRight/components/SearchSection/SearchSection.tsx
+++ b/src/features/singleRight/components/SearchSection/SearchSection.tsx
@@ -27,7 +27,7 @@ import {
   ServiceStatus,
   type ServiceWithStatus,
 } from '@/rtk/features/singleRights/singleRightsSlice';
-import { debounce } from '@/resources/utils';
+import { arraysEqual, debounce } from '@/resources/utils';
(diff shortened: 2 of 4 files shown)


src/components/Filter/OptionDisplay.tsx
```diff
@@ -89,7 +89,7 @@ export const OptionDisplay = ({
     return (
       <button
         className={classes.option}
-        key={option.value}
+        key={option.label}
```
Albertlarsen, line 92: because resource register returns both Arbeids and velferdsetaten and Nav (Arbeids og velferdsetaten)

allinox: Does that mean we will now have both Arbeids and velferdsetaten and Nav (Arbeids og velferdsetaten) in our display?
That sounds suboptimal...

allinox (changes requested):
I still don't understand what caused that "kasse" situation in the bug report, but I can't trigger it when I run this branch locally, so I guess it's fixed somehow..? 😅

src/components/Filter/OptionDisplay.tsx
Albertlarsen: I don't see any other solution than checking if 2 resourceOwners have the same org code and then only return one of them (We could then add a rule for which of them to return in those scenarios but I don't know what rule we should use for which of them to use, maybe we should huddle about it tomorrow?)

Albertlarsen:
@allinox Sorry, I should have said something about it. The reason of the weird Kasse error was because there where duplicate keys in react. "Arbeids and velferdsetaten" and "Nav (Arbeids og velferdsetaten)" had the same key since they had the same org code. Duplicate keys in react causes unexpected errors and this was one of them (Seemed like everytime you searched for something "Arbeids and Velferdsetaten" was automatically added as a duplicate and also always displayed, maybe because it didn't actually exist but was just a duplicate). So duplicate keys caused more duplicate keys and so on so everytime you searched for something, more duplicate keys was added and more duplicate FilterOptions were added.

src/features/singleRight/components/SearchSection/SearchSection.tsx
```diff
@@ -272,7 +272,6 @@ export const SearchSection = ({ onAdd, onUndo }: SearchSectionParams) => {
           values={filters}
           onApply={(filters) => {
             setFilters(filters);
-            setCurrentPage(1);
```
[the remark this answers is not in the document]
Albertlarsen: Good and well thought through comment :)

src/components/Filter/OptionDisplay.tsx
allinox: Yeah, no I don't think we should try and fix every issue that isn't our fault with special case code.
That being said though,, I thought the way the code was written in the BFF was that we filtered out the orgs from the list of resources, based on the orgcode or orgnumber? So shouldn't that have meant that we wouldn't get duplicates like these?
(I haven't actually cheched right now though, so I might be wrong)

allinox:
@Albertlarsen 
Aha! That makes sense!
Thanks for the explanation and kudos for figuring it out! 🙌

allinox (approved):

Issue #788, opened 2024-03-12
Apply use of landmark roles


allinox:
![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/48394f2e-2615-4a20-b082-e26a5280b57d)
WCAG rule 1.3.1 requires each section of each page to be semantically described in the code, using (among other things) landmarks: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#3._landmark_roles

Other than headers, we hardly use landmarks, and we should use them more.

The available landmarks are:
- [banner](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role) (document [`header`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header))
-  [complementary](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/complementary_role) ([`aside`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside))
-  [contentinfo](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/contentinfo_role) (document [`footer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer))
-  [form](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/form_role) ([`form`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form))
-  [main](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/main_role) ([`main`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main))
-  [navigation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role) ([`nav`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav))
-  [region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role) ([`section`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section))
-  [search](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role) ([`search`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search))

[...]

PR #928, opened 2024-06-03
Chore/general/788-apply-use-of-landmark-roles

sonwit:
Apply use of landmark roles + a bit of refactoring

## Description
- main-tag added to page component
- header-tag moved to page-header. [Mdn docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header)
- search-tag on search-forms + search results
- use grid for layout to remove duplicated code 
- fix heading levels so they are in correct order

## Related Issue(s)
- #788

## Verification
- [x] All tests run green

## Documentation


src/components/PageContainer/PageContainer.tsx
```diff
@@ -22,7 +22,7 @@ export const PageContainer = ({ children }: PageContainerProps) => {
         <div className={classes.closeButtonContainer}>
           <button
             className={classes.closeButton}
-            aria-label={String(t('common.close'))}
+            aria-label={String(t('common.close'))} // Should this rather be "Avslutt" or "Cancel" as you are not closing anything?
```
allinox, line 25: Yeah, I see your point here.. 🤔
I think this is inherited from the mentality in A2 where this page was actually a modal. These x-es are usually more associated with closing widows than anything else and I guess you still _could_ argue that you're closing the process you're in, but yeah.. 
I don't actually have a very strong opinion on the matter personally. "Cancel" is probably just as understandable, and probably technically more correct 😊

src/features/apiDelegation/offered/ChooseOrgPage/ChooseOrgPage.tsx
```diff
@@ -110,147 +107,92 @@ export const ChooseOrgPage = () => {
...
+          <RestartPrompter
+            spacingBottom
+            restartPath={
+              '/' + ApiDelegationPath.OfferedApiDelegations + '/' + ApiDelegationPath.ChooseApi
+            }
+            title={t('common.an_error_has_occured')}
+            ingress={t('api_delegation.delegations_not_registered')}
+          />
+        )}
         <PageContent>
```
allinox, line 127: Moving this outside of the pageContent makes the page look a bit strange imo
The gap of blue in the middle makes the page look broken, and the shadow that is still present as a border on that gap adds to that feeling.

The regular content is now also rendered unlike before, and while I'm not necessarily opposed to the principle of that idea, I wonder if it was intentional and whether you've taken into abbount if the info in the info-alert and the page content could contradict eachother and confuse the user?

image

src/features/singleRight/delegate/ChooseServicePage/ChooseServicePage.tsx
```diff
@@ -101,12 +102,12 @@ export const ChooseServicePage = () => {
                     onAdd={onAdd}
                     onUndo={onRemove}
                   />
-                  <GroupElements>
+                  <div className={classes.navigationContainer}>
```
allinox, line 105: I can only see that the new NavigationSection component is used on the (so far unused) request version of this page. Is there a reason why it's not used here as well?

allinox (changes requested):
Great work again! 😄
Especially the adding of grid templates saves us a lot of duplicate code and space-heavy logic so thank you for coming up with this solution instead! 🙌

The only reason why I'm not approving this right away is because moving the restart prompter on the APIdelegation page makes the page look a bit broken. I would love to discuss that further 😊

allinox (approved):
Great work! 😄
Love the new NavigationSection!

Issue #909, opened 2024-05-10
Clear out delegableOrg slice


allinox:
With the use of rtkQuery most of this file can be moved to other places for better state handling

- [x] Move lookupOrg dispatch function to rtkQuery in lookupApi
- [x] Move chosenDelegableOrgList from state in delegableOrg and place it in apiDelegationSlice (as chosenOrgs)
- [x] Move choice logic to ChooseOrgPage
- [x] Move searchInCurrentOrgs to ChooseOrgPage (utils file?)
- [x] Delete delegableOrgSlice

PR #936, opened 2024-06-20
Remove delegableOrgSlice

allinox:
## Description

This is a technical dept PR that cleans up several aspects:
- Clears out delegableOrgSlice:
  - Moves lookupOrg dispatch function to rtkQuery in lookupApi
  - Moves chosenDelegableOrgList from state in delegableOrg and place it in apiDelegationSlice (as chosenOrgs)
  - Moves choice logic to ChooseOrgPage
  - Moves the search to a separate hook
  - Deletes delegableOrgSlice
- Fixes broken component tests
- Fixes github action workflow so running of component tests are triggered by changes in src (as opposed to ignoring changes to src as it appears to have been mistakenly set to)

## Related Issue(s)
- #909 

## Verification
- [ ] All tests run green

## Documentation


src/features/apiDelegation/offered/ChooseOrgPage/useOrgSearch.tsx
allinox, line 1: I'm not sure if this file should be placed here because it is only used on this page, or if it should be in the hooks-folder like the others. What do you think @sonwit?

src/features/apiDelegation/components/OverviewPageContent/OrgDelegationActionBar/OrgDelegationActionBar.tsx
```diff
@@ -60,7 +60,7 @@ export const OrgDelegationActionBar = ({
             color={'second'}
             size={'medium'}
             onClick={softRestoreAllCallback}
-            aria-label={String(t('common.undo')) + ' ' + organization.orgName}
+            aria-label={String(t('common.undo')) + ' ' + organization.name}
```
sonwit, line 63: Should this include a reference to the action you are undoing? Something like: "undo delete orgName"?  

for example: aria-label={`{t('common.undo')} {t('api_delegation.delete')} {organization.name}`}

sonwit: And also, should we inform the user that the item will be deleted only after a confirm has been clicked?

sonwit: Not important(!) and not really related to this PR, so it might be that this is a task that needs its own issue in the backlog? IDK... 🤷🏼

src/rtk/features/apiDelegation/apiDelegationSlice.ts
sonwit, line 1: This is much simpler! Love it 🫶

src/features/apiDelegation/components/OverviewPageContent/OrgDelegationActionBar/OrgDelegationActionBar.tsx
allinox: > Should this include a reference to the action you are undoing? Something like: "undo delete orgName"? 
> for example: aria-label={{t('common.undo')} {t('api_delegation.delete')} {organization.name}}
Great idea! I'll put it in right now (don't really see any reason to wait)

> And also, should we inform the user that the item will be deleted only after a confirm has been clicked?
Also a good idea, but I'm not sure exactly what is the best way to do so.
Just add a line that straight up tells them there's a save button they need to click on? Or is there a more elegant way to do it?
What do you think?

src/features/apiDelegation/offered/ChooseOrgPage/useOrgSearch.tsx
sonwit: I thing keeping it where it is makes sense for now, given its specialized nature for the org-search component. It's good to be mindful of potential reusability in the future, and if we do end up needing the search functionality elsewhere, i think refactoring it into a reusable component in a "common" folder would be a great way to go.

src/features/apiDelegation/components/OverviewPageContent/OrgDelegationActionBar/OrgDelegationActionBar.tsx
sonwit: Maybe we could change the text to something more descriptive like "add/remove {orgname} to delete list" or something like it?

allinox: Yeah, but that's very programmer-specific language. Maybe "mark <the thing> for deletion"?
Or we could maybe just move the user from the button to the save button if you click on it to make it clear that you have to save? Or find some other way to inform the user that you need to save for this decision to come into effect?

I mean, that's what we do for the seeing users - inform them of the need to save by making a save button pop up and thus making them infer that they have to press it.

sonwit: I definitely agree "mark for deletion" is a more intuitive term.

I'm not so sure about moving the focus to the save button. Users may lose track of where they were in the interface, and if the user needs to make multiple changes before saving, having the focus repeatedly jump to the Save button can become annoying and inefficient.

Maybe we could use a live region to announce changes in the page content dynamically. After the user interacts with the delete button, the live region could announce "Changes made. Click the Save button to apply."

allinox: Yeah, I think the live region might be the best solution.
Then we can announce that "<the thing> is set to be deleted, click the save button to save this change"

allinox: Or yeah, "Changes made. Click the Save button to apply" is even easier!

allinox: Done! 😊

sonwit (approved):
Awesome! The ApiDelegationSlice is a great improvement. ❤️ 
Also, super fix on  the aria-label thing! 
Our visually impaired users will be singing your praises.

src/features/apiDelegation/components/OverviewPageContent/OrgDelegationActionBar/OrgDelegationActionBar.tsx
sonwit: Nice! 👍🏼

PR #954, opened 2024-06-28
New route: Users

allinox:
## Description

This PR adds a new route to the ones already in use, where the new GUI will be built.
It adds a changing background color, different from the one in production

In addition to this some of the mock data is supplemented or changed for the users/parties to more accurately reflect the data in AT22 for a more seamless transition from real to mock

## Related Issue(s)
- #941 

## Verification
- [ ] All tests run green

## Documentation


src/components/PageWrapper/PageWrapper.module.css
```diff
@@ -0,0 +1,4 @@
+.background {
```
allinox, line 1: This is a suggestion for dealing with dynamically changing backgrounds, but I'm not sure it's the best one.
I see that inbox manually sets the same background color on each component instead and to me that seems a bit tedious, but idk - what do you think?

sonwit: This looks like a good approach.
I agree that using CSS variables here is a solid option. It helps centralize style definitions, and its easy to replace later if we have to!

src/features/singleRight/delegate/ReceiptPage/ActionBarSection/ActionBarSection.tsx
```diff
@@ -120,7 +119,6 @@ export const ActionBarSection = ({ recipientName }: ActionBarSectionProps) => {
           <Ingress
             className={cn(extraSpacing && classes.successText)}
             spacing
-            level={2}
```
sonwit, line 123: Nice 🫶

sonwit (approved):
This is a great start! I really like that you've implemented CSS variables. 
It's also great that you've been able to reuse our existing translation files, route setup, and build pipeline. This definitely saves us a lot of time and effort.
Looking forward to seeing how this develops!

Issue #967, opened 2024-07-04
API-delegation - ChooseOrgPage crashes


allinox:
### Description of the bug

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/501b0d5c-64a2-4d7c-9fb2-dc2391b6e1ce)

This bug was introduced very recently, with the transformation from UserInfoSlice to UserInfoApi.

Seems like a reference to the old slice state was overlooked in the process.

To do:
- [ ] Switch to using the new rtkquery on ChooseOrgPage (marked in the image underneath)
- [ ] Go through the solution and check that there are no other references to the old state
- [ ] Check that both the APIDelegation and Singleright delegation flow works alright

![image](https://github.com/Altinn/altinn-access-management-frontend/assets/22012997/3c1c0eca-a439-4906-9f8a-0a340c77d7c5)

### Steps To Reproduce

1. Go to one of the AT-environments
2. Log in and go thtough the API-delegation flow
3. Observe that the solution crashes on the choose org page

PR #968, opened 2024-07-04
Replace redux-userinfo with rtq userInfoApi

sonwit:
## Description
Fix page crash & use new  userInfoApi instaed of userinfo from redux.

## Related Issue(s)
- #967

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/rtk/features/userInfo/userInfoApi.ts
+++ b/src/rtk/features/userInfo/userInfoApi.ts
@@ -6,10 +6,6 @@ interface UserInfoApiResponse {
     name: string;
   };
 }
-interface ReporteeApiResponse {
-  organizationNumber: any;
-  name: string;
-}
 
 interface UserInfo {
   name: string;
@@ -17,7 +13,7 @@ interface UserInfo {
 
 interface ReporteeInfo {
   name: string;
-  orgNumber?: string;
+  organizationNumber?: string;
 }
 
 export const userInfoApi = createApi({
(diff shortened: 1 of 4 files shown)


src/rtk/features/userInfo/userInfoApi.ts
```diff
@@ -25,18 +31,21 @@ export const userInfoApi = createApi({
...
-    getReportee: builder.query<UserInfordata, void>({
+    getReportee: builder.query<ReporteeInfoData, void>({
       query: () => `reporteelist/${getCookie('AltinnPartyId')}`,
       keepUnusedDataFor: 300,
       transformResponse: (response: ReporteeApiResponse) => {
-        return { name: response.name };
+        return {
+          name: response.name,
+          orgNumber: response.organizationNumber,
+        };
```
allinox, line 48: Also nitpic, but I don't think you need to transform the response when you just want to return it as the object specified in the query definition.
This queary will automatically return a ReporteeInfoData object as is.

TransformResponse is for when you wish to return something else, or process the data before returning it. (See singleRightApi for examples without transformResponse)

allinox (approved):
Thanks for being so quick to fix this! 😄
Just added some suggestions that don't really impact the functionality

src/rtk/features/userInfo/userInfoApi.ts
```diff
@@ -7,13 +7,19 @@ interface UserInfoApiResponse {
...
+  organizationNumber: any;
   name: string;
 }
 
-interface UserInfordata {
+interface UserInfoData {
   name: string;
 }
 
+interface ReporteeInfoData {
```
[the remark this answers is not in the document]
sonwit: I agree. I have changed the variables to just use "info" instead of both "info" and "data" 👍🏼

src/rtk/features/userInfo/userInfoApi.ts
sonwit: Great tip 👍🏼 . I removed the "transformResponse" ang changed the code to use organizationNumber instead of org-number!

PR #969, opened 2024-07-05
Adding a list components to show users on the "/users" route.

sonwit:
Adding list component to show users on the users-page

## Description
I have implemented the initial version of the list component for the users page. This component aims to display users in a list format and serves as a foundation for further development. Additionally, I have started on the list-item component, which is also included on the users page. Currently, it is a basic implementation designed to provide a visual reference for future enhancements.
Feel free to modify the components as needed. I hope this implementation will serve as a starting point as we identify additional requirements for the rest of the page.

## Related Issue(s)
- #942
- #943 (started)
- #944 (started)

## Verification
- [ ] All tests run green

## Documentation
Skjermbilde 2024-07-05 kl  15 37 47


src/components/List/List.module.css
```diff
@@ -0,0 +1,39 @@
...
+  border-radius: 0.5rem;
+}
+
+.listItem {
+  background-color: #fff;
+  list-style-type: none;
+  padding: 0;
+  margin: 0;
+  display: flex;
+  flex-direction: column;
```
mgunnerud, line 30: listItem currently has only one child, so flex-direction is not needed (perhaps not the other flex styling either)

src/components/UserIcon/UserIcon.module.css
```diff
@@ -0,0 +1,23 @@
...
+  margin: 0.5rem;
+}
+
+.userIcon::before {
+  content: attr(data-user-initials);
+  position: absolute;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 2.5rem;
```
mgunnerud, line 14: The width and height for the absolute positioned element should be `100%`, so we can change the sizes by changing the parent values 😊

src/components/UserIcon/UserIcon.module.css
```diff
@@ -0,0 +1,23 @@
...
+  height: 2.5rem;
+  border-radius: 50%;
+  border: #23262a 2px solid;
+  background-color: #fff;
+  color: #23262a;
```
mgunnerud, line 19: add a variable for border and background color, since they have the same value?

src/components/UserIcon/UserIcon.module.css
```diff
@@ -0,0 +1,23 @@
+.userIcon {
+  display: block;
```
mgunnerud, line 2: .userIcon should have `postion: relative` so the ::before element is positioned according to the position of the element with .userIcon class

src/components/UserIcon/UserIcon.module.css
```diff
@@ -0,0 +1,23 @@
...
+  width: 2.5rem;
+  height: 2.5rem;
```
mgunnerud, line 4: Add a variable for width and height? These should probably always be the same value

src/features/amUI/users/UsersList.tsx
```diff
@@ -0,0 +1,86 @@
...
+        ) : (
+          <>
+            <UserIcon userName={user.name} /> {user.name}
+          </>
+        )}
+        {user.role && <Tag>{user.role}</Tag>}
+      </div>
+    </ListItem>
+  );
+};
```
mgunnerud, line 86: Should we move this hook into the \<UserListItem\> component, and let that component handle it's own expanding? The component also receives `user` as a prop, so the expanded \<List compact\> markup can also be handled by the component.

We should also add a `key` to \<UserListItem /\> 😊 (should also be added to testdata)

```suggestion
       {users.map((user) => (
        <UserListItem
          key={user.name}
          user={user}
        >
      ))}
    </List>
  );
};

const UserListItem = ({ user }: { user: UserType }) => {
  const [isExpanded, setIsExpanded] = useState(false);
  const isExpanable = !!(user.users && user.users.length > 0);

  return (
    <>
      <ListItem>
        <div className={classes.listItemContent}>
          {isExpanable ? (
            <Button
              onClick={() => setIsExpanded((oldExpanded) => !oldExpanded)}
              variant='tertiary'
              icon
            >
              {isExpanded ? (
                <ChevronUpCircleIcon fontSize={'3.3rem'} />
              ) : (
                <ChevronDownCircleIcon fontSize={'3.3rem'} />
              )}
              {user.name}
            </Button>
          ) : (
            <>
              <UserIcon userName={user.name} /> {user.name}
            </>
          )}
          {user.role && <Tag>{user.role}</Tag>}
        </div>
      </ListItem>
      {isExpanded && user.users && (
        <List compact>
          {user.users.map((user) => (
            <UserListItem key={user.name} user={user} />
          ))}
        </List>
      )}
    </>
  );
};
```

src/components/List/TestDataUserList.ts
```diff
@@ -0,0 +1,53 @@
+export interface UserType {
+  name: string;
+  role?: string;
+  users?: UserType[];
+}
+
+export const testUsers: UserType[] = [
+  {
+    name: 'Dan Daglig Leder',
```
mgunnerud, line 9: We should probably add an id field to each user, which we can use as unique identifier in our code 😊

mgunnerud (changes requested):
Great initial version! I have added some comments which I believe could make the component slightly easier to modify later 😊

allinox (commented):
Nice! You even started on the listItems themselves and even the userIcons! 😄

This is a great starting point for our page - already starting to look like the design.
To speed things up a bit, I hope it's ok if someone else continues with this branch while you're on vacation. Maybe Martin, if he has the time?
He has a lot of good comments 😊

Enjoy your summer! ☀️

src/components/List/List.tsx
```diff
@@ -0,0 +1,22 @@
...
+  compact,
+}: {
+  children: React.ReactNode[];
+  heading?: ReactNode;
+  compact?: boolean;
+}) => {
+  return (
+    <>
+      {heading}
+      <ul className={classNames(classes.list, compact ? classes.compact : classes.spacious)}>
```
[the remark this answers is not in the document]
mgunnerud: I tried to change the list to `List` from design system, but that list includes some extra styling (for example bottom-margin for each \<li\> which we would need to add more css to work around 😅). I suggest we use our own \<ul\> + \<\li\> implementation for now

allinox (approved):
Looks great! 😄

I would probably make the ListItem a more fleshed-out component along the lines of the ActionBar (which can then be either expandable or not) so that the UserList and other listing areas can just use that component outright without much extra logic but I'm thinking that can come later in its own task.

For now, this is a fantastic start! 🙌

src/components/List/List.tsx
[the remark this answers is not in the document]
allinox: Sound reasoning - I agree 😊

mgunnerud (approved):

PR #975, opened 2024-07-09
Present users as a paginated list (with mocked data)

allinox:
image

## Description

This PR adds logic for presenting a list of right holders (someone holding rights/power of attorney for the organization) in a paginated manner. It contains no styling but does contain the following:
- Added new BFF endpoint for fetching right holders. This is used by the react frontend to fetch right holders
- Mocking of right holder data
  - Static data is added as it can be beneficial to have static data that behaves as expected, both in manual and integration testing
  - Dynamically generated mock data for testing of performance on large number of right holders. Will be especially interesting when search and filter is added
  - Users for accessing these two different types of mock data have been added/updated and a guide has been added to the repos readme
- Reusable pagination logic has been added to the react application and used to list out the right holders with navigational functions for the pages 
- BFF mocking has been enabled for AT22. This env will thus be running on mocked data both for this new amUI solution and the previous packages singleRights and apiDelegation

## Related Issue(s)
- #945 

## Verification
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green


src/resources/utils/arrayUtils.tsx
```diff
@@ -8,3 +13,39 @@ export function arraysEqual<T>(array1?: T[], array2?: T[]): boolean {
...
+ * Calculates the total number of pages needed to paginate an array.
+ * @param array The array to be paginated.
+ * @param numPerPage Number of elements on each page.
+ * @returns The total number of pages.
+ */
+export function getTotalNumOfPages<T>(array: T[], numPerPage: number): number {
+  if (numPerPage <= 0) {
+    throw new RangeError('numPerPage must be greater than 0');
+  }
+  return Math.ceil(array.length / numPerPage);
```
mgunnerud, line 50: It will probably not happen for the user list, but if a paged array has no items, this function will return 0. If the user list has no items, the `getArrayPage` function will throw `new RangeError('Invalid page number');`. Should we return total number of pages === 1 when an array has no items?

src/resources/utils/arrayUtils.tsx
```diff
@@ -8,3 +13,39 @@ export function arraysEqual<T>(array1?: T[], array2?: T[]): boolean {
...
+ * @param numPerPage Number of elements on each page.
+ * @returns The elements present on the given page.
+ * @throws {RangeError} Thrown when either numPerPage or page is not valid (e.g., asking for a page that does not exist).
+ */
+export function getArrayPage<T>(array: T[], page: number, numPerPage: number): T[] {
+  if (numPerPage <= 0) {
+    throw new RangeError('numPerPage must be greater than 0');
+  }
+
+  if (page <= 0 || page > Math.ceil(array.length / numPerPage)) {
```
mgunnerud, line 30: Can we reuse `getTotalNumOfPages` here? 😄

```suggestion
  if (page <= 0 || page > getTotalNumOfPages(array, numPerPage)) {
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/UserService.cs
```diff
@@ -48,5 +50,24 @@ public async Task<AuthorizedParty> GetPartyFromReporteeListIfExists(int partyId)
...
+            {
+                PartyUuid = rightHolder.PartyUuid,
+                PartyType = rightHolder.Type,
+                Name = rightHolder.Name,
+                PersonId = rightHolder.PersonId,
+                OrganizationNumber = rightHolder.OrganizationNumber,
+                UnitType = rightHolder.UnitType,
+                RegistryRoles = rightHolder.AuthorizedRoles.Where(role => Enum.IsDefined(typeof(RegistryRoleType), role.ToUpper()))
+                    .Select(role => (RegistryRoleType)Enum.Parse(typeof(RegistryRoleType), role.ToUpper()))
+                    .ToList()
```
mgunnerud, line 69: I noticed that the fake data can add the same role several times, can we add `.Distinct()` to this query to avoid duplicate roles?
```suggestion
                RegistryRoles = rightHolder.AuthorizedRoles.Where(role => Enum.IsDefined(typeof(RegistryRoleType), role.ToUpper()))
                    .Select(role => (RegistryRoleType)Enum.Parse(typeof(RegistryRoleType), role.ToUpper())).Distinct()
                    .ToList()
```

mgunnerud (approved):
Great work! 😀 I added some minor comments

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/UserService.cs
allinox: Good point!
However, I think that is more a flaw of the automatically generated test data as no user will have multiples of the same role in prod (unless something is wrong)
How about I add the distinctiveness to the data generation instead?

src/resources/utils/arrayUtils.tsx
allinox: Let's do it! 😄

src/resources/utils/arrayUtils.tsx
allinox: Good catch! We absolutely should!
Adding that and a test to it now! 👍

Issue #860, opened 2024-04-16
Memoization bug


Albertlarsen:
### Description of the bug

<img width="438" alt="image" src="https://github.com/Altinn/altinn-access-management-frontend/assets/31471142/0379e334-1596-4169-9039-88a60fedf7ee">

### Steps To Reproduce

1. Go to chooseService page in SingleRights

PR #983, opened 2024-07-23
Bugfix: calls to useAppSelector should return the same result

mgunnerud:
## Description

Any manipulation of data should be done outside `useAppSelector` call, so the call to `useAppSelector` will always return the same data based on the same input

## Related Issue(s)
- #860 

## Verification
- [ ] All tests run green

## Documentation


allinox (approved):
Thank you for fixing this long-standing bug! 😄

PR #987, opened 2024-07-24
Søk i brukerliste

mgunnerud:
## Description

This PR adds a simple user search, based on the assumption that all users (and organizations) are available to the frontend code. Is is possible to search by:
- Name (case insensitive, partial search)
- Organization `organizationNumber` (exact search)
- user `personId` (exact search)

For search matches in users in `inheritingRightHolders`, only the search matches will be displayed when the organization is expanded.

Filter button or logic is not added by this PR
 
## Related Issue(s)
- #949 

## Verification
- [ ] All tests run green

## Documentation


src/features/amUI/users/UsersList.tsx
```diff
@@ -42,6 +81,9 @@ export const UsersList = () => {
           </Heading>
         }
       >
+        {pageEntrees.length === 0 && searchString && (
```
sonwit, line 84: Should this be "entries" instead of "entrees"?

src/features/amUI/users/UsersList.tsx
```diff
@@ -42,6 +81,9 @@ export const UsersList = () => {
...
+          <div>{t('users_page.user_no_search_result')}</div>
```
sonwit, line 85: To inform users with screen-readers  of a update on the page, could we add a live region here to inform them that no matches were found?

src/features/amUI/users/UsersList.tsx
```diff
@@ -24,12 +32,43 @@ export const UsersList = () => {
     if (!rightHolders) {
       return [[], 1];
     }
-    const numPages = getTotalNumOfPages(rightHolders, pageSize);
-    return [getArrayPage(rightHolders, currentPage, pageSize), numPages];
-  }, [rightHolders, currentPage]);
+
```
sonwit, line 35: Nice implementation of the search! Its clear and concise, and it works great! 
We can consider optimizing with something like inverted indexes if we need to later if performance becomes an issue.

src/features/amUI/users/UsersList.tsx
```diff
@@ -24,12 +32,43 @@ export const UsersList = () => {
...
+    setCurrentPage(1); // reset current page when searching
+  };
 
   return (
     <div className={classes.usersList}>
+      <Search
+        className={classes.searchBar}
+        placeholder={t('users_page.user_search_placeholder')}
+        value={searchString}
+        onChange={(event) => onSearch(event.target.value)}
```
sonwit, line 69: Is it a good idea to debounce this? Continuous updates to the results list as the user types might be visually distracting and resource-intensive on the browser.

allinox: I agree. We use debouncing on the other searches and even if this one doesn't call the BFF, there's no reason to use more resources than necessary

src/features/amUI/users/useDebouncedFilteredRightHolders.ts
```diff
@@ -0,0 +1,76 @@
...
+  const isOrgNrMatch = rightHolder.organizationNumber === searchString;
+  return isNameMatch || isPersonIdMatch || isOrgNrMatch;
+};
+
+export const useDebouncedFilteredRightHolders = (
+  searchString: string,
+  currentPage: number,
+  pageSize: number,
+) => {
+  const { data: rightHolders, isLoading: isFetching } = useGetRightHoldersQuery();
```
allinox, line 18: Not important, but in rtkQuery, isFetching is actually a possible return which tells us if the queary is loading either to get the data the first time or on a refetch, while isLoading tells us if the query is loading the data for the very first time. I'm assuming you are intentionally using isLoading here, but could maybe renaming the variable to isFetching potentially cause confusion for later reading of the code?

src/features/amUI/users/useDebouncedFilteredRightHolders.ts
```diff
@@ -0,0 +1,76 @@
...
+  const computePageEntries = () => {
+    if (!rightHolders) {
+      setPageEntries([]);
+      setNumOfPages(1);
+      setSearchResultLength(0);
+      setLoading(false);
+      return;
+    }
+
+    const filteredRightHolders: RightHolder[] = searchString ? [] : rightHolders;
```
allinox, line 33: Since there is a possibility of filter being added later, maybe this should be called searchResults or something instead?

src/localizations/no_nb.json
allinox, line 1: Should there not been added any nynorwegian texts?

allinox (changes requested):
Great work! 😄
Readable and the functionality seems to be working swell as far as I can see.
I just have a few suggestions. Let me know if there's anything you want to discuss.

src/features/amUI/users/UsersList.tsx
```diff
@@ -1,64 +1,81 @@
...
-        totalPages={numOfPages}
-        onChange={(newPage) => setCurrentPage(newPage)}
-        nextLabel='Neste'
-        previousLabel='Forrige'
-      />
+        role='alert'
+        spacing
+      >
+        {searchResultLength === 0 ? t('users_page.user_no_search_result') : ''}
+      </Heading>
```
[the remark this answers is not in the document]
sonwit: Dynamically adding a DOM element that has aria-live or a role that has an implicit aria-live is typically not honored by the screen reader. Keeping it in the dom is a way to make sure the user gets the information we put in it. [link](https://stackoverflow.com/questions/76849643/best-practice-to-add-role-alert-in-react-conditional-rendering)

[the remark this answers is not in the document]
allinox: True, but I get Wave errors (accessibility) on the empty heading.
Maybe the live region can instead be a div and then the content (the heading) can be rendered conditionally?
Or we can switch from a heading to a paragraph instead? Then it will no longer be an empty heading-error and as far as I know there is no rule against empty paragraphs

src/features/amUI/users/UsersList.tsx
```diff
@@ -1,64 +1,81 @@
...
+        className={classes.searchBar}
+        placeholder={t('users_page.user_search_placeholder')}
+        value={searchString}
+        onChange={(event) => onSearch(event.target.value)}
+        onClear={() => onSearch('')}
+      />
       <List
         compact
         heading={
           <Heading
```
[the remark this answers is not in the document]
sonwit: Sure. Maybe we can use aria-labelledby to link the list to the heading above?

src/features/amUI/users/UsersList.tsx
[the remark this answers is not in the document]
sonwit: Sure. A paragraph is a good option. I made it a heading as it is what NAV does, but it doesnt have to be a heading i think :)

src/features/amUI/users/UsersList.tsx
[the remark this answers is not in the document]
allinox: Good idea!

src/features/amUI/users/useDebouncedFilteredRightHolders.ts
sonwit: The loading-prop is not in use, so i remove t for now.

src/localizations/no_nb.json
sonwit: i fix 👍🏼

src/features/amUI/users/useDebouncedFilteredRightHolders.ts
sonwit: 👍🏼

allinox (approved):
Search completed ✅
Wonderful work from you both! 🕵️🕵️‍♂️

PR #1002, opened 2024-08-08
Display logged in user on top of the page (and create UserItem component)

allinox:
image

## Description

This PR creates a new component to be used when displaying the users on page in a sized hierarchy.
It uses this to display the logged in user on top of the page, extracting it from the list of other right holders.

In addition to this there is a slew of other (minor) changes such as:
- Making the content list of buisness right holders more visibly part of an hierarchy
- Adding aria-expanded and aria-controls to the items that expands and linking them to their content for accessibility pourposes
- Hiding roles on inherining right holders (considering displaying them in a different way than tag in the future)
- Added hover-colors to clickable elements
- Fixing a bug with the generated test data that made all inherited parties have the same name
- Fixed bug where mocked data of Livsglad Film had the wrong userId for AT22
- Other, even more minor things

## Related Issue(s)
- #947 

## Verification
- [ ] All tests run green

## Documentation


src/components/UserIcon/UserIcon.module.css
```diff
@@ -11,5 +11,20 @@
   color: #ffffff;
   text-align: center;
   text-transform: uppercase;
-  margin-right: 1rem;
+}
+
+.lg {
+  --userIcon-size: 2.5rem;
```
sonwit, line 17: Nice!. smart use of css-variables. 

Not important, but is the "userIcon-size" on line 2 necessary? 
if it needs to be there, the default size is "md", so maybe it should be 2rem by default here as well?  

`.userIcon {
  --userIcon-size: 2.5rem;
  min-width: var(--userIcon-size);`

sonwit (approved):
Great work!

Nice cleanup of the user-list file.
Nice to have a reusable list-item that handles multiple sizes for the future.
Nice implementation of the "extract user" function!

I just had one comment, but I don't know if it's worth changing.

src/components/UserIcon/UserIcon.module.css
allinox: I'm not sure if it's needed actually... 🤔
I'm just used to variable always having to be declared before they're used 😅

But you're absolutely right! It should at least be set to the default value.
I'll fix it tomorrow 😊

PR #1006, opened 2024-08-12
Add expand/collapse based on serch-result

sonwit:
## Related Issue(s)
- #998 
- 
## Verification
- [ ] All tests run green

## Documentation


allinox (approved):
Great work! 🙌
The search results look more related to the search string now 😄

Another way of triggering the opening could be to instead expand on the UserItem implementation so that it's expanded state can be (optionally) set externally or to be rendered with an initial expanded state. But unless we see the need for that in other cases as well, adding a state to RightHolder is probably fewer lines of code.

PR #1052, opened 2024-08-29
Display popular services on SingleRights

allinox:
image

## Description
Added logic to display a set of 10 popular services instead of the full unfiltered list of services on an empty search.
This is only enabled in tt02 and prod (where these services actually are) and is done to hide the many test services from view unless actually searched for.

## Related Issue(s)
- #1033

## Verification
- [ ] All tests run green

## Documentation


backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/appsettings.json
```diff
@@ -25,5 +25,8 @@
     "Issuer": "amui",
     "App": "access-management-ui",
     "CertificateName": "JWTCertificate"
+  },
+  "LogicFlags": {
+    "DisplayPopularSingleRightsServices": false
```
allinox, line 30: Despite what we talked about earlier, this ended up kinds feature-flaggy anyway... 😅
I tried giving it a name that didn't scream feature flag though, if that makes sense, but maybe that's a mistake too? What do you think?

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceService.cs
```diff
@@ -57,9 +61,20 @@ public async Task<PaginatedList<ServiceResourceFE>> GetPaginatedSearchResults(st
...
 
-                List<ServiceResourceFE> filteredresources = FilterResourceList(resourcesFE, resourceOwnerFilters);
-                List<ServiceResourceFE> searchResults = SearchInResourceList(filteredresources, searchString);
-                return PaginationUtils.GetListPage(searchResults, page, resultsPerPage);
+                bool displayPopularServicesOnly = _logicFlags.DisplayPopularSingleRightsServices;
+                if (string.IsNullOrEmpty(searchString) && (resourceOwnerFilters == null || resourceOwnerFilters.Length == 0) && displayPopularServicesOnly)
+                {
+                    // Return a selection of popular services
+                    List<ServiceResourceFE> popularResources = FilterOutPopularResources(resourcesFE);
+                    return PaginationUtils.GetListPage(popularResources, page, resultsPerPage);
```
allinox, line 69: I'm not sure how to write this into a test when its based on environment variables.. Any suggestions?
Or isn't it needed do you think?

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/appsettings.json
sonwit: Yeah, this seems like a good solution for now!
Since it's essentially controlling the visibility of a feature, it might be clearer to just embrace the term "feature flag" and rename it accordingly.

I still think we probably need the ability to quickly enable/disable the functionality without requiring code deployments in the future. but we can implement that at a later time. 
👍

src/features/singleRight/components/SearchSection/SearchSection.tsx
```diff
@@ -62,6 +63,10 @@ export const SearchSection = ({ onAdd, onUndo }: SearchSectionParams) => {
     resultsPerPage: searchResultsPerPage,
   });
 
+  const env = getEnvFromUrl();
+  const displayPopularResources =
+    !searchString && filters.length === 0 && (env === Environment.PROD || env === Environment.TT02);
+
```
sonwit, line 69: I'm not entirely convinced that this is the best approach. Currently, the logic for turning features on or off is both in the bff and the client. I propose centralizing this logic to make it easier to manage and maintain. 

One potential solution is to pass the feature flags from the backend for frontend (BFF) in the `window` object. This would allow us to access them directly from the client-side code.

We can achieve this by:

1. **Create a `featureFlags` object:**

   ```csharp
   if (await ShouldShowAppView())
   {
       ViewBag.featureFlags = _logicFlags; // _logicFlags or feature_flags?
       return View();
   }
   ```

2. **Serialize and embed in the view:**

   ```cshtml
   <body>
     <script type="text/javascript">
       window.featureFlags = @Html.Raw(Json.Serialize(ViewBag.featureFlags));
     </script>
     <div id="root"></div>
   ...
   </body>
   ```

3. **Use the feature flags in js:**

   ```javascript
   const displayPopularResources = window.featureFlags.displayPopularResources;
   ```

I believe this approach provides better maintainability and separation of concerns. However, I'm curious to hear what you think.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceService.cs
sonwit: Hm. I'm not sure how we could do this. Maybe we can override the value in "Logic-Flags" somewhere, but I'm not entirely sure how we could do that. It's probably not the most important thing to test, so I don't think we need to worry about it at the moment.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI/appsettings.json
allinox: I agree on both counts - I'll rename this and then we can add a more sophisticated solution later 😊

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/ResourceService.cs
allinox: I'm thinking the same thing.
If this becomes a recurring problem with feature-flags and the likes, we can take a look at that point

src/features/singleRight/components/SearchSection/SearchSection.tsx
allinox: This is so much better!
I absolutely agree with you that one single source of truth is better than two and for the BFF to share it's flags with the frontend is a great way to go about it (that I wasn't aware could be done like this).
Thank you for this excellent feedback! 😄🙌

sonwit (approved):
Nice! This implementation looks good, and it's great to see feature flags being used. This will definitely make it easier to manage the rollout and any future changes.

PR #1064, opened 2024-09-10
Feat/ny brukerflate/1035 enkelttjenester på brukerside

sonwit:
Showing delegated single-rights on user-page 

## Description
- Add GUI for single-rights on user-page
- add mock data for getting single-rights 
- add test for getting delegated single-rights for user

## Related Issue(s)
- #1035 

## Verification
- [ ] All tests run green

## Documentation


src/resources/hooks/usePagination.ts
allinox, line 1: We actually already have Util functions for pagination in the arrayUtils file (used among other places on the users page to paginate the users list). I don't know if it is better to use a hook than a simple util function in this case, but we should probably stick to one of the two.

Some things to consider: Do you have any preference for hooks for cases like these? Do we need to store the currentPage in a state? And do we need the nextPage, prevPage, and gotToPage functions for pagination?

allinox (approved):
Great job with this! So clean and neat-looking! 😄
Nothing really show-stopping in terms of QA, but I have a couple of notes and suggestions I'd like to hear your thoughts on 😊

src/localizations/en.json
```diff
@@ -205,7 +205,9 @@
     "user_no_search_result": "No users found"
   },
   "user_rights_page": {
-    "page_title": "User access - Altinn"
+    "page_title": "User access - Altinn",
+    "single_rights_title": "Single rights",
```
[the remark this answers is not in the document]
sonwit: Yes. I tottaly agree. Services is much better.

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Tests/Controllers/SingleRightControllerTest.cs
```diff
@@ -303,6 +310,58 @@ public async Task CreateDelegation_AltinnApp_valid()
...
+
+            // Assert
+            Assert.Equal(expectedResponse.Count, actualResponse.Count);
+            foreach (var right in expectedResponse)
+            {
+                var actual = actualResponse.FirstOrDefault(r => r.Identifier == right.Identifier);
+                Assert.NotNull(actual);
+                Assert.Equal(right.Title, actual.Title);
+                Assert.Equal(right.Description, actual.Description);
+            }
```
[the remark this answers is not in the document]
sonwit: Absolutely! Thanks for pointing that out.
I'll update this right away! Appreciate the tip!

src/resources/hooks/usePagination.ts
sonwit: Sounds good, thanks for the feedback!

I agree we should definitely stick to using the arrayUtils for pagination to keep things consistent. I'll update the hook to use those util functions instead.

You're right about not needing nextPage and prevPage, I'll remove those. However, I think we need to keep currentPage in state for calculating which items to display.

I generally think that a hook is generally a better approach for separating and reusing common logic like this, as it involves handling state. It seems like a more flexible and reusable solution compared to the hook used in useFilteredRightHolders for example.

Looking forward to hearing your thoughts on the updated hook!

allinox: I agree!

This way, we don't have to keep the current-page state in the separate components, but can instead keep it in one reusable hook.
Well argumented and well done! 🙌

PR #1081, opened 2024-09-20
Feat: Delegate new service (new GUI)

allinox:
## Description

What this PR does:
- Adds new trigger button to right holder page that opens a modal for delegating new single rights
- Makes the modal and content for the simplified happy-flow of delegation:
  - Searchable services are presented and clickable
  - A second view for the modal to display description and specific rights
  - This two-staged flow ends with the delegation of the chosen rights. (which works with actual backend as well, using existing APIs)
  - (Further description of the flow can be found in the related issue)
- Moving the used calls from slice to query. Note that though this new implementation uses these new queries, the old slice functions are not removed due to their use in the old GUI that is currently in production (and changing the old GUI implementation is out of scope for this PR)
- Updated UserIcon component to new Avatar component for more reusability. (This will eventually be exchanged with the external Avatar component shared with Arbeidsflate)
  - Note: For now, service owner avatars are just file icons since we currently lack a way to get their logos
- Added Intelligent Albatross to mocked list of right holders as this user already has mock data for single rights delegation
- Made minor temporary changes to right holder's page to fit into the new display of already delegated services.

## Related Issue(s)
- #1037 

image
image
image


src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,41 @@
...
+  /** If displaying an icon istead of initials */
+  icon?: React.ReactNode;
+  /** Determines the shape of the avatar */
+  profile?: AvatarProfile;
+  /** Additional class names for styling */
+  className?: string;
+}
+
+export const Avatar = ({ icon, name, size = 'md', profile = 'person', className }: AvatarProps) => {
+  console.log(name?.charAt(0));
```
sonwit, line 23: console.log

sonwit (approved):
Great work on the delegation modal. I only have a few comments, so I'll approve this right away :D.

The modal seems to render a little larger than the window on sizes between 1100px and 530px, but it works great otherwise :).

Skjermbilde 2024-09-23 kl  15 03 39

src/components/Avatar/Avatar.tsx
allinox: 🙈

PR #1090, opened 2024-09-25
inherit secrets needed for tt02 release pipe

Nyeng:
Needed because Playwright 

## Documentation


allinox (approved):
🙌

PR #1100, opened 2024-09-30
Fix SRR error message

allinox:
## Description

Changed to new text for more understandable srr error message

The way error texts are handled here is not ideal and the texts need some clean-up, but since this solution is dying anyway, I'll leave it as it is for now, and then we can implement a better way of doing things and remove the unused texts when this solution is removed as well.

## Related Issue(s)
- #1098 

## Verification
- [ ] All tests run green

## Documentation


sonwit (approved):
Nice fix. Better error messages for the users! 👍

PR #1121, opened 2024-10-18
fix typescript errors + rebuild yarn.lock file

sonwit:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/rtk/features/singleRights/singleRightsSlice.ts
+++ b/src/rtk/features/singleRights/singleRightsSlice.ts
@@ -283,7 +283,7 @@ const singleRightSlice = createSlice({
           if (sws.service?.identifier === serviceID) {
             const delegationResponseList: Right[] = [
               {
-                rightKey: `${action.meta.arg.resourceReference[0].id}:${action.meta.arg.resourceReference[0].value}`,
+                rightKey: ``,
                 resource: action.meta.arg.resourceReference,
                 action: '',
                 status: errorCode,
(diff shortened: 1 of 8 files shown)


src/rtk/features/singleRights/singleRightsSlice.ts
```diff
@@ -282,6 +283,7 @@ const singleRightSlice = createSlice({
           if (sws.service?.identifier === serviceID) {
             const delegationResponseList: Right[] = [
               {
+                rightKey: `${action.meta.arg.resourceReference[0].id}:${action.meta.arg.resourceReference[0].value}`,
```
allinox, line 286: I don't think we're the ones that define the right key 🤔
We should be getting it from the backend directly, I think, so that we don't risk creating different and invalid rightKeys (since we're going to be using them when informing backend of what rights to delegate in the future)

allinox: Nvm, I see now that this is for the rejected case.
I guess it doesn't really matter in that case since we can't delegate the rights anyway

allinox (approved):

src/rtk/features/singleRights/singleRightsSlice.ts
sonwit: Good point. Ill remove it anyway... We can pass in an empty string to prevent typescript from complaining...

PR #1130, opened 2024-10-24
User can edit actions on delegated service

allinox:
## Description

This PR adds functionality for editing already given services. User can now click on an already given service/the power of attorney for a service, see what actions are given there and make changes.

In order to make this, several things have been done:
- BFF:
  - Changed endpoint for deleting to match the new and coming backend endpoints, splitting it into one for deleting an entire service delegation (that is all actions) and another for deleting only a single right/action.
  - Created new edit endpoint which makes use of the right deletion endpoint and the already existing (but possibly to be outdated eventually) endpoint for delegation. 
  - Note that we don't yet know what the new backend endpoints will return, so this PR makes simplified assumptions which can then be edited when we know more.
  - Added tests for new endpoints.
  - Changed mock data to make it easier to work with and test editing as a feature
  - Added temporary (hard-coded to single user) cookie `AltinnPartyUuid` from HomeController for easier access to the reportee's uuid, which will soon replace partyId as the main id used for referencing parties. A task for removing this temporary solution has been added in #1126, which should be done when Altinn 2 has started setting this cookie on its own.
- React:
  - Created new hooks for delegation, deletion, and edit to abstract away complexities and help reduce complexity/size of individual components
  - Created reusable delete button component which is used on both the overview and in the modal for viewing a specific delegation.
  - Added an rtkQuery to get the full reportee party information as a shorthand.
  - Added logic to display what actions are active and not on already delegated services.
  - Cosmetic styling changes for ListItems to display hover state and other minor things.

**Note that the service A3 app 2 cannot be edited the way other resources might be.** When clicking the single chip, it will not appear unchecked, yet the save button implies there are changes to be saved. This represents an issue with the current display for apps and the forced boolean representation of a multifaceted state. If a user is given some but not all of the delegable actions on an app, clicking the chip will toggle the delegated rights off and the not delegated rights on. There is currently no way of editing an A3 app other than deleting the whole thing and then re-delegating it.
We should consider not displaying the edit button on A3 apps, but for now, I think it makes for a neat display of the problems with the current simplification.

## Related Issue(s)
- #1038 

Already delegated resources are marked as such in search for resources:
image

The same view is used both for inspecting a given resource directly from the users page and in the modal for making new delegations (apart from the back button which takes you back to the search view)
image
image

## Verification
- [ ] All tests run green

## Documentation


sonwit (approved):
This PR looks great! I really like the new from/to/resource format, it makes things much clearer.

The issue with editing A3 apps is a problem we need to fix. I agree that keeping the edit button for now is a good way to highlight the problem.

I approve and have no comments. Great work! 👍

PR #1134, opened 2024-10-30
Force reload when user has changed reportee in another tab

allinox:
## Description

This PR introduces a check on the value of the cookie `AltinnPartyId` (which later will be changed to AltinnPartyUuid) so as to force a reload if the value changes due to a change in reportee in a different tab.
The user is informed that the reload is to be happening with an alert.

This will hopefully solve the issue of possibly delegating something on behalf of a different party than intended which could potentially happen if the reportee is changed in one tab while performing a delegation in another without the GUI being updated.

## Related Issue(s)
- #1133 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/resources/Cookie/CookieMethods.ts
+++ b/src/resources/Cookie/CookieMethods.ts
@@ -16,21 +16,27 @@ export function getCookie(cname: string) {
   return '';
 }
 
+/**
+ * Detects changes in a given cookie by polling it with the given interval
+ * @param cookieName The name of the cookie to be checked
+ * @param interval The interval in which to poll the cookie (given in milliseconds)
+ * @returns `true` if the cookie has been changed from it's original value, `false` otherwise.
+ */
 export const useCookieListener = (cookieName: string, interval = 2000) => {
-  const [cookie, setCookie] = useState(getCookie(cookieName));
-  const [displayAlert, setDisplayAlert] = useState(false);
+  const [cookieValue, setCookieValue] = useState(getCookie(cookieName));
+  const [cookieChanged, setCookieChanged] = useState(false);
 
   useEffect(() => {
     const checkCookie = setInterval(() => {
       const currentCookie = getCookie(cookieName);
-      if (currentCookie !== cookie) {
-        setCookie(currentCookie);
-        setDisplayAlert(true);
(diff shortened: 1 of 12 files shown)


src/resources/Cookie/CookieMethods.ts
```diff
@@ -13,3 +15,22 @@ export function getCookie(cname: string) {
...
+      if (currentCookie !== cookie) {
+        setCookie(currentCookie);
+        setDisplayAlert(true);
+      }
+    }, interval); // Check every interval
+
+    return () => clearInterval(checkCookie);
+  }, [cookie, cookieName]);
+
+  return displayAlert;
```
sonwit, line 35: I'm not sure if this will be used for anything else than displaying an alert, but in the context of a "useCookieListener"-function, should "displayAlert" also have a more generic name like "cookieChanged" or "cookieUpdateDetected"?

src/resources/Cookie/CookieMethods.ts
```diff
@@ -13,3 +15,22 @@ export function getCookie(cname: string) {
...
+    const checkCookie = setInterval(() => {
+      const currentCookie = getCookie(cookieName);
+      if (currentCookie !== cookie) {
+        setCookie(currentCookie);
+        setDisplayAlert(true);
+      }
+    }, interval); // Check every interval
+
+    return () => clearInterval(checkCookie);
+  }, [cookie, cookieName]);
```
sonwit, line 33: I suggest not having the cookie in the dependency array. We use setCookie inside the interval, and whenever the cookie state changes, the effect will re-run, potentially causing issues. 

```suggestion
  }, [cookie, interval]);
```

sonwit (approved):
Nice work! I left a few comments. 

I'll approve this and you can merge when you are ready. 👍🏼

src/resources/Cookie/CookieMethods.ts
allinox: Good point - I'll fix! 👩‍🔧

src/resources/Cookie/CookieMethods.ts
allinox: I agree - I'll change it 🫡

PR #1144, opened 2024-11-07
Fixed missing scopes in API delegation overview

allinox:
## Description

Noticed that scopes were no longer showing in the overview, so made a quick fix.

Also removed attempt at bypassing the CAPTCHA in tt02 that ended up not working.


sonwit (approved):
👍

Issue #70, opened 2022-12-20
Received API-accesses | S2 | ReceivedApiAccessesOverview


Albertlarsen:
### Description

Reuse overview component and create ReceivedApiAccessesOverview

<img width="582" alt="image" src="https://user-images.githubusercontent.com/31471142/208639918-9d0af233-d7e6-4619-9792-5826f4e66082.png">

PR #1169, opened 2024-11-25
Display access package delegations

allinox:
image

## Description

This PR introduces a display of which access packages that a given user has received.
The packages are sorted into areas and displayed as either an access given or one that is not, in two separate lists. The ability to manipulate these lists are to come later.

- A new endpoint is introduced to the BFF and another has been outlined in comments underneath.
- Obs: A lot of models have been introduced in this PR but a lot of them are somewhat tentative as changes in backend models might result in changes here too. So far, this is a mix of best guesses and wishes.
- The view of the delegated packages have been split into multiple components in order to keep the size and complexity of each file to a minimum

**Note: The users that have been mocked to have access packages are Intelligent Albatross and Sitrongul Medaljong**

## Related Issue(s)
- [#70](https://github.com/Altinn/altinn-authorization-tmp/issues/70)

image


backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Models/AccessPackage/AccessPackage.cs
```diff
@@ -20,6 +22,11 @@ public class AccessPackage
         /// </summary>
         public string Description { get; set; }
 
+        /// <summary>
+        /// Description
```
sonwit, line 26:
```suggestion
        /// Resources
```

PR #1184, opened 2024-12-06
Can delete access package delegation from info modal

allinox:
image

## Description

This PR introduces revoking/deleting of access packages, enabling it first on the info view of the access package modal.

- A new endpoint has been added to the BFF. As the return model from the backend is still uncertain and as the scope of this task only concerns whether or not the revoking failed in the most rudimentary terms, minimal effort has been put into the return model.
- Regret-functionality has been scoped out of this task as it is not currently supported by the snack bar and because it is uncertain whether or not it is even necessary in the info modal case as the delete button is immediately transformed into a delegate button upon deletion, which has the same functionality as a regret button anyway.
- A bug was discovered, where the invalidation of tags and subsequent refetching of delegated access packages triggered closing of the info modal when not in search mode. This has now been fixed.
- Also discovered that refetching delegations closed all open areas. This would likely cause issues later when adding delegation and deletion from the overall view, and so a fix was implemented. The expanded state in the AccessAreas is now controlled externally, where areas that have been opened are remembered between renders.

## Related Issue(s)
- #[73](https://github.com/Altinn/altinn-authorization-tmp/issues/73)

## Verification
- [ ] All tests run green

## Documentation


backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/IAccessPackageService.cs
```diff
@@ -23,5 +24,14 @@ public interface IAccessPackageService
         /// <param name="languageCode">The code of the language on which texts are to be returned</param>
         /// <returns>A dictionary of lists (sorted by access area) containing all access package delegations that the right holder has on behalf of the specified right owner</returns>
         Task<Dictionary<string, List<AccessPackageDelegation>>> GetDelegationsToRightHolder(Guid rightHolderUuid, Guid rightOwnerUuid, string languageCode);
+
+        /// <summary>
+        ///     Performs a search for access packages based on the provided parameters and sorts them into a list of areas for frontend to display
```
sonwit, line 29:
```suggestion
        ///     Revokes access to a given package for a right holder (to) on behalf of a party (from)
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/AccessPackageService.cs
```diff
@@ -75,5 +75,11 @@ public async Task<Dictionary<string, List<AccessPackageDelegation>>> GetDelegati
 
             return sortedAccesses;
         }
+
+        /// <inheritdoc />
+        public Task<HttpResponseMessage> RevokeAccessPackage(Guid from, Guid to, string packageId)
+        {
+            return _accessManagementClient.RevokeAccessPackage(to, from, packageId);
```
sonwit, line 82: Screenshot 2024-12-09 at 11 05 59

It looks like the from and to parameters have swapped places here? 

```suggestion
            return _accessManagementClient.RevokeAccessPackage(from, to, packageId);
```

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/Interfaces/IAccessPackageService.cs
allinox: Whoops! 🙈

sonwit (approved):
Nice job on implementing revoking of access packages!

The bug fixes for tag invalidation and refetching delegations are much appreciated!

It looks like there was a small mixup with the params in AccessPackageService. I also noticed a minor copy-paste error.

This looks good overall! Approving. 👍🏼

backend/src/Altinn.AccessManagement.UI/Altinn.AccessManagement.UI.Core/Services/AccessPackageService.cs
allinox: Great catch! 🙌
I switched midway through to always having from first to keep it more consistent with the path in the bff endpoint, but I must have forgotten the Service interface.
Thanks for noticing! 😄

PR #1185, opened 2024-12-09
Display error message on failed Access List delegation check

allinox:
image

## Description

Added support for displaying error message when delegation check fails due to resources using access lists. The displayed message is the same as for srr-errors for A2-services

## Related Issue(s)
- #1123

image


sonwit (approved):
🙌🏼

PR #1198, opened 2024-12-19
Enhance BFF mock to enable running with mock data only on new GUI

allinox:
## Description

This PR is a big one, and makes a lot of changes to how the clients and mock clients are structured for fetching and editing data in access management.

The client for access management has been split in two: One for the old am api and another for the new am api to be used only in the new GUI. The former of these two is marked as V0. Details on what has been done is described more in detail in the related issue

The mock is now enabled on a client-based level, meaning that it is up to the developer exactly which client they want to run mocked when running locally, whether it be a full mock or partial one. In AT22, we will run mock on the am client and the access packages client, while all else is running with the true endpoints. This is now also set as default in development, but anyone can feel free to change this. (Running the resource registry locally as well can often be a boon too, since this allows us more control over which cases we can test locally without having to look for test data in at22)

AT23, AT24, TT02 and PROD is set to run without any mocking at all.

## Related Issue(s)
- #1156 

## Verification
- [ ] All tests run green

## Documentation


README.md
```diff
@@ -162,17 +162,23 @@ Happy coding!
...
 
-The mock can be activated locally or in a specific environment by going to the relevant appsettings file and setting `UseMockData` to true. Be aware that this should NEVER be done for TT02 or PROD!
+The mock can be activated locally or in a specific environment by using the appsettings file for that environment and setting the relevant flags in the 'MockSettings' section. Each flag here relates to a client in the BFF and thus, to the integration to external backends. Setting one of these to true means that instead of the regulat client being used at runtime, a mocked version will be used instead. The mocked client will use data from static files instead of making calls to external sources. Since each client can be mocked individually, there is a number of possible combinations that can be enabled, depending on the needed case. In development, you are free to use whatever combination you want, and in certain situations, we enable mocking in one or more of the AT environemts. That being said...
```
sonwit, line 165:
```suggestion
The mock can be activated locally or in a specific environment by using the appsettings file for that environment and setting the relevant flags in the 'MockSettings' section. Each flag here relates to a client in the BFF and thus, to the integration to external backends. Setting one of these to true means that instead of the regular client being used at runtime, a mocked version will be used instead. The mocked client will use data from static files instead of making calls to external sources. Since each client can be mocked individually, there is a number of possible combinations that can be enabled, depending on the needed case. In development, you are free to use whatever combination you want, and in certain situations, we enable mocking in one or more of the AT environments. That being said...
```

allinox: Ahh, yes, it wouldn't have been an Alex-PR without the regularly scheduled typos 😛

sonwit (approved):
Great work on implementing this! This is a significant improvement to how we handle access management data for testing. 
Awesome work on refactoring the clients and implementing the new mocking system. It will be super exciting to show this around and get feedback, as well as seeing it in play with the inbox app.🤩
I've had a look through the code, and everything looks good to me! 👍  No obvious issues. 🙌
I approve!☃️✨

PR #1248, opened 2025-01-29
Fix formatting on dete resource snackbar

allinox:
image

## Description

Formatting is no longer an issue when the resource name is not printed through i18next.
In the long run we should reevaluate whether or not to even use snackbars for feedback in the case of deletion. At least when the operation fails we should probably open the infomodal instead to display a more detailed description of why the operation failed

## Related Issue(s)
- https://github.com/Altinn/altinn-access-management-frontend/issues/1191

## Verification
- [ ] All tests run green

## Documentation


sonwit (approved):
👍🏼

PR #1255, opened 2025-02-03
Feat/amui/204 info modal for admin roller

sonwit:
## Description
- Added new methods for role delegation, revoke and search (this is used for getting all roles, i am unsure if the backend will support search for roles, but i assume some kind of filter is needed, so i made it a search)

- Added DelegationCheck for roles as we need to know if a role can be delegated.

- Added DetailCode enum in BFF for error codes on delegation-check.

- Added new properties to the Role and RoleAssignment models such as IsDelegable, ProviderId, Area, and Inherited. After discussing with backend, this reflects how the data will look when we get data from the actual backend. 

- Updated frontend components to support new role functionalities.

## Related Issue(s)
- #[204](https://github.com/Altinn/altinn-authorization-tmp/issues/204)
- #[205](https://github.com/Altinn/altinn-authorization-tmp/issues/205)

## Verification
- [ ] All tests run green

## Documentation


allinox (approved):
Great work on what turned out to be a large task! 👏

I like the trend we're on in striving to make several smaller components instead of 400-line monstrosities, and you continue that beautifully here.
Great work also with the screen reader testing and adding aria-titles for the otherwise invisible divide of the two lists: Active roles and available roles.

I've made some suggestions for improvements, which you can consider at your leisure. Otherwise - approved! 🚀😄

PR #1316, opened 2025-03-06
Feat/amui/1299 update ds to v1.0

sonwit:
update ds to v1.0
- size & color -> data-size & data-color
- Override design-tokens for "neutral-surface" and default-button-color
- various fixes for things that broke with the update

## Related Issue(s)
- #https://github.com/Altinn/altinn-access-management-frontend/issues/1299

## Verification
- [ ] All tests run green

## Documentation


allinox (approved):
I went through our solution to check that it looked ok 🔍
I only found some minor things, but I just fixed them as I went.
Otherwise, it looks good in my book! Some colors are a bit off, but I think that's because of the mismatch between the component library and design system, so I'm thinking those will fix themselves once Inge has fixed the colors in the component library.

Great job with this! 😄🙌

PR #1400, opened 2025-04-04
Add OldRolesAlert component

sonwit:
## Description
Adding alert to inform users that we are updating Altinn with a link to the old altinn solution

I followed the design sketches, but I’m not sure if I agree with the location and the text. I’d love to hear your thoughts.

I also changed the link text to make it a bit more self-explanatory outside of the previous paragraph. 
 
 
## Related Issue(s)
- #[208](https://github.com/Altinn/altinn-authorization-tmp/issues/208)

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- **New Features**
  - Introduced an "Old Roles Alert" notification that now appears across several access and rights pages to inform users about legacy role authorizations.
  - Updated on-screen text for role descriptions and alerts, including new messaging about the Altinn renewal, to provide clearer and more inclusive guidance.
  - Added a new `SnackbarProvider` context for improved notifications in the `AccessPackageList` component.
  - Created new Storybook stories for the `OldRolesAlert` and `RoleList` components to enhance development documentation.

- **Bug Fixes**
  - Removed outdated role description from the user rights page to simplify user information.

### DIFF
--- a/src/features/amUI/common/OldRolesAlert/OldRolesAlert.tsx
+++ b/src/features/amUI/common/OldRolesAlert/OldRolesAlert.tsx
@@ -1,13 +1,20 @@
-import { getHostUrl } from '@/resources/utils/pathUtils';
 import { useTranslation } from 'react-i18next';
 
 import { ExternalLinkIcon } from '@navikt/aksel-icons';
 import { Link } from 'react-router';
 import { Alert, Heading, Paragraph, Link as DsLink } from '@digdir/designsystemet-react';
 import styles from './OldRolesAlert.module.css';
+import { usePartyRepresentation } from '../PartyRepresentationContext/PartyRepresentationContext';
+import { getRedirectToSevicesAvailableForUserUrl } from '@/resources/utils';
+import { useFetchRecipientInfo } from '@/resources/hooks/useFetchRecipientInfo';
 
 export const OldRolesAlert = () => {
   const { t } = useTranslation();
+  const { toParty, fromParty, selfParty } = usePartyRepresentation();
+
+  const { userID, partyID } = useFetchRecipientInfo(toParty?.partyUuid ?? '', null);
+
+  const url = getRedirectToSevicesAvailableForUserUrl(userID, partyID);
   return (
     <Alert data-color='info'>
       <div className={styles.container}>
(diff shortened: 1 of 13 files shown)


src/localizations/no_nb.json
```diff
@@ -345,6 +344,11 @@
     "trigger_button": "Legg til bruker",
     "add_button": "Legg til bruker"
   },
+  "a2Alerts": {
+    "launchAlertHeading": "Vi fornyer Altinn",
+    "launchAlertContent": "Denne brukeren kan også ha fullmakter fra gamle Altinn.",
+    "launchAlertLinkText": "Se fullmakter fra gamle Altinn"
+  },
```
allinox, line 351: I'm not sure I like the formulations here either, since I feel like we're just saying the same thing twice without explaining _why_ the accesses aren't the same.
But we can probably figure that out later. For now, I was just wondering if we maybe shouldn't call Altinn2 for "old", but rather "classic", like github does when they introduce new features?

image

src/features/amUI/common/OldRolesAlert/OldRolesAlert.tsx
```diff
@@ -0,0 +1,37 @@
...
+            marginBottom: 'var(--ds-size-2)',
+          }}
+        >
+          {t('a2Alerts.launchAlertHeading')}
+        </Heading>
+
+        <Paragraph>{t('a2Alerts.launchAlertContent')}</Paragraph>
+        <DsLink asChild>
+          <Link
+            to={`${getHostUrl()}ui/profile/`}
```
allinox, line 27: As a courtesy to the user, I think we should redirect them directly to the rights-page of the user they are currently looking at, instead of just to the profile page. That would save them the trouble of having to find the same user in the A2-list of users and then opening their page.

This is something we already do in the old single-rights GUI, so you can actually get the url just using this function: `redirectToSevicesAvailableForUser` 😊

allinox (approved):
Nice work adding the info alert and adding more stories ✨
I agree with you, in that I'm not super happy with the placement of the alert either.
You could maybe consider moving it over the heading inside the tab, but the placement and formulation could probably be something we should discuss with UX in general anyway, so you can also just leave it be for now 😊

src/localizations/no_nb.json
sonwit: Good point. Ill update the text. 👍🏼

PR #1401, opened 2025-04-04
Minimize other available packages in area

allinox:
image

## Description

I went to the Designsystem channel and asked for advice about expandables inside expandables, and they suggested using a simple button to show and hide the extended list instead, which is something that Brønnøysundregistrene uses on their page. This PR is a test for this case.
The argument from BR was that as long as the button itself explains what it should do and the content being shown is close to the trigger, there is no need for the use of expanded and aria-controls, since the user will be able to figure it out from the context.

A button is thus added to toggle the display of the available packages in the area.
An aria-label is added to the chevron in the button, indicating whether the button will open or close the display of these other packages. Aria-labels are also added to the lists themselves to indicate the difference between them.

The number of available packages is displayed in the button for clarity

## Related Issue(s)
-  https://github.com/Altinn/altinn-authorization-tmp/issues/501

## Note
The packages are not minimized inside the delegation modal
image

## Summary by CodeRabbit

- **New Features**
  - Added new options to control the display of available access packages, allowing users to toggle or minimize package listings.
  
- **Style**
  - Introduced a new button style to improve the presentation of the package display option.
  
- **Documentation**
  - Enhanced user interface text by adding new labels and titles for access package sections in multiple languages.

### DIFF
--- a/src/features/amUI/common/AccessPackageList/AreaItemContent.tsx
+++ b/src/features/amUI/common/AccessPackageList/AreaItemContent.tsx
@@ -1,20 +1,21 @@
-import { Alert, Heading, Paragraph, Spinner } from '@digdir/designsystemet-react';
+import { Alert, Button, Heading, Paragraph } from '@digdir/designsystemet-react';
 import { ListBase } from '@altinn/altinn-components';
 import { useTranslation } from 'react-i18next';
 import React, { useMemo, useState } from 'react';
+import cn from 'classnames';
+import { ChevronDownIcon, ChevronUpIcon } from '@navikt/aksel-icons';
 
 import type { AccessPackage } from '@/rtk/features/accessPackageApi';
+import { useAccessPackageDelegationCheck } from '@/resources/hooks/useAccessPackageDelegationCheck';
+import type { ActionError } from '@/resources/hooks/useActionError';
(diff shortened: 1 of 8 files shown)


src/features/amUI/common/AccessPackageList/AreaItemContent.tsx
```diff
@@ -93,8 +85,38 @@ export const AreaItemContent = ({
           ))}
         </ListBase>
       )}
-      {packages.available.length > 0 && (
-        <ListBase>
+
+      {!showAvailablePackagesExternal && (
+        <Button
```
sonwit, line 90: Kunne vi legge på aria-expanded=false + aria-controls="<liste-id>" ref [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded#buttons)?  
Tror det ville gjøre det tydligere for skjermlesere at denne listen kan ekspanderes...

sonwit: Sorry. Så at du kommentrete det PRen din. Bare ignorer komentaren 🤦🏼

sonwit (approved):
Nice job on hiding the 'other packages'. It looks much cleaner now. And great that you checked with the design system team how to use expandables inside expandables 
✅

PR #1411, opened 2025-04-10
Make scrollbar less noticable

allinox:
image

## Description

To avoid the content moving to adjust for suddenly appearing scrollbars, (most notably on windows machines and on the rights page of a person when opening an access area or when opening and closing a modal), a scrollbar-gutter has been added.
Since styling the scrollbar has become a lot easier lately, I also added made it thinner and added a bit of color to make it blend more seamlessly into the background. (This is also more notable on windows, where the default scrollbar is a real eyesore 👇)

image

One problem with this approach, however, is that the scrollbar-gutter shows up on top of the designsystem overlay, as a white bar at the edge of the screen. So far, I have not been able to fix this issue, but I would love to hear your view on it, and whether this issue is a dealbreaker. Or, pershaps if you think all this styling is too much hassle to even bother with at all.

image

## Related Issue(s)
- https://github.com/Altinn/altinn-authorization-tmp/issues/538

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- **Style**
  - Enhanced scrollbar appearance across the interface with a thinner design.
  - Updated scrollbar styling for various components, including service lists and search results.
  - Improved color schemes for scrollbar tracks and thumbs for a more refined, modern look.

### DIFF
--- a/src/resources/css/Common.module.css
+++ b/src/resources/css/Common.module.css
@@ -5,13 +5,14 @@ body {
 html {
   scrollbar-gutter: stable;
   scrollbar-width: thin;
+  background-color: var(--ds-color-background-subtle);
   scrollbar-color: var(--ds-color-neutral-border-subtle) rgba(0, 0, 0, 0);
 }
 
 :root {
   font-family: 'Inter', sans-serif;
   --backgroundColorA2: #1eadf7;
-  --backgroundColorA3: #e9f5ff;
+  --backgroundColorA3: var(--ds-color-background-subtle);
 
   --a2-semantic-surface-action-first-no_fill: #00315d;
 
--- a/src/resources/css/Common.module.css
+++ b/src/resources/css/Common.module.css
@@ -2,6 +2,12 @@ body {
   margin: 0;
 }
 
+html {
+  scrollbar-gutter: stable;
+  scrollbar-width: thin;
+  scrollbar-color: var(--ds-color-base-default) var(--ds-color-background-subtle);
+}
+
 :root {
   font-family: 'Inter', sans-serif;
   --backgroundColorA2: #1eadf7;
--- a/src/resources/css/Common.module.css
+++ b/src/resources/css/Common.module.css
@@ -5,7 +5,7 @@ body {
 html {
   scrollbar-gutter: stable;
   scrollbar-width: thin;
-  scrollbar-color: var(--ds-color-base-default) var(--ds-color-background-subtle);
+  scrollbar-color: var(--ds-color-neutral-border-subtle) rgba(0, 0, 0, 0);
 }
 
 :root {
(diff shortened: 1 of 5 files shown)


src/resources/css/Common.module.css
```diff
@@ -2,6 +2,12 @@ body {
   margin: 0;
 }
 
+html {
+  scrollbar-gutter: stable;
+  scrollbar-width: thin;
+  scrollbar-color: var(--ds-color-base-default) var(--ds-color-background-subtle);
```
sonwit, line 8: Screenshot 2025-04-11 at 11 57 29

I notice that adding background-color here, the scrollbar in the tabs goes a little crazy. Maybe setting the background color on the html elemt itself is a better solution? 
```suggestion
  scrollbar-color: var(--ds-color-base-default); 
  background-color: var(--ds-color-background-subtle);
```

src/resources/css/Common.module.css
```diff
@@ -2,6 +2,12 @@ body {
   margin: 0;
 }
 
+html {
+  scrollbar-gutter: stable;
```
sonwit, line 6: Also, when we open a dialog, the backdrop doesn't fill the entire screen. It looks a bit weird on small screens especially. It might be a better solution to add a permanent scrollbar, or add a condition to this so that this styling only applies to larger screens. 
```suggestion
 overflow-y: scroll;
```
Screenshot 2025-04-11 at 11 51 46

allinox: Yeah, that's the problem I was talking about in the PR description, that I haven't been able to find a solution for 😅
The permanent scrollbar is an interesting thought. However, in my experience, the scrollbar automatically disappears when the dialog opens, leaving only the gutter behind. So I'm not sure if always having a scrollbar there will make a difference....

Oh, but, it doesn't show up on iPad and mobile view for me, though (when in simulated touch screens). Is that just me? 🤔

image

sonwit: Yes. sorry. Its probably correct that it dosn't show up on mobile screens... so this might wont be an issue. I was testing with firefox on descktop.

src/resources/css/Common.module.css
allinox: Good catch!
What do you mean with the html element itself? 🤔 
As far as I could see (though, I'm not really an expert on this,) the scrollbar for the entire page is not set by us, but belongs to the `html`-tag that encompasses our code, so the only place I was able to set a different color than the default, was to set it here.

Also, I don't see this artifact either when using the mobile simulator, only when making the desktop screen thin. Is it just me? 😅

allinox: That being said, I'm not very happy with the approach of setting the color for the full page and then having to override it everywhere else, so I would love it if you had a suggestion of another alternative!

sonwit (approved):
Great work on updating the styling on the scrollbars. They look much better now, especially on Windows.
I think it would be better if we moved the scrollbar background to the html element, but otherwise it looks good.
We also might consider moving this into the css we share with "arbeidsflate" in the future to ensure consistent behavior, but we can keep it here for now. 

✅

PR #1464, opened 2025-06-03
Added cascading delete on revoke of rightholder and chenged rightholder cashing

allinox:
## Description

This PR adds cascading delete on revoking of rightholders, meaning that if a rightholder has an access (such as an access package) they now lose this package during the revoke.

Also, the time that unused data for rightholders is cached has been changed from 300 seconds to only 3 seconds, so that changes such as deleting a rightholder will be updated almost instantaneously in the GUI.

Invalidation of tags is not used here since we need a couple of seconds to play an animation before navigating the user back to the users page and invalidation of cache causes refetching also on currently used data such as the check that the user is able to view the page of the party (which they can only do for parties with connections ...and this connection was just revoked. Meaning: an error occurs)

## Related Issue(s)
- https://github.com/Altinn/altinn-authorization-tmp/issues/799

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- **Improvements**
  - Enhanced revocation of right holders to include cascading behavior.
  - Improved data freshness for right holder information by reducing cache duration.
  - Optimized data updates for right holders in the user information section.


sonwit (approved):
👍

PR #1496, opened 2025-06-23
Add validation error message when delegation fails

sonwit:
## Description
This PR returns the relevant error codes for failed delegations and adds them to the error message.
- We currently only handle all cases of "not allowed" as one. We will hopefully be able to expand this with more granular validation messages soon.
- Hopefully this can be used as a framework to build upon as we handle more cases for validation and error messages.

## Related Issue(s)
[872](https://github.com/Altinn/altinn-authorization-tmp/issues/872)

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/features/amUI/common/DelegationModal/AccessPackages/AccessPackageInfo.tsx
+++ b/src/features/amUI/common/DelegationModal/AccessPackages/AccessPackageInfo.tsx
@@ -68,6 +68,7 @@ export const AccessPackageInfo = ({ accessPackage, availableActions = [] }: Pack
     return null;
   }, [activeDelegations, isFetching, accessPackage.id]);
 
+  const { displayLimitedPreviewLaunch } = window.featureFlags;
   const userHasPackage = delegationAccess !== null;
   const accessIsInherited =
     (delegationAccess &&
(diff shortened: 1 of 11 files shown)


src/features/amUI/common/DelegationModal/AccessPackages/AccessPackageInfo.tsx
```diff
@@ -169,6 +171,11 @@ export const AccessPackageInfo = ({ accessPackage, availableActions = [] }: Pack
                 size='xs'
                 status={actionError.httpStatus}
                 time={actionError.timestamp}
+                message={
```
allinox, line 174: I am wondering if maybe we shouldn't display the status and timestamp in the cases where there isn't a technical error, since the http status won't actually be of use for the user and may end up distracting from the actual message because it looks more like a technical error (even if it is not). 🤔
Removing it also makes the error smaller, which is probably a plus.
Maybe either add a bool to the technical error component to hide the technical details, or not use the TechnicalParagraph component for non-technical errors, and just display the validation error sentence alone inside the alert.
image

allinox (approved):
Another round of great work! 😄
Particularly in using the problem-factory approach to mapping the error codes 🙌

PR #1733, opened 2025-11-19
Feat/role inherited statusmessage

sonwit:
The role info modal is now showing a message explaining where the access comes from.
I have also added some stories to the Storybook. 

Inheritance Cases:
**via_role**: 
the permission contains both viaRole and via, meaning the user inherits this role because of another role they hold in the originating organization.

**via_parent**: 
the permission only has via, and the acting party is the same as the to party, so the child unit inherits the role from its parent organization.

**via_agent**: 
the permission only has via, and the acting party is the same as the from party, so the user acts as an agent and inherits the role through the intermediary unit listed in via.

To test the different cases: 
Via Role: 
 - log in with (05887798518) and choose "Blaut Økonomisk Isbjørn SA" as the actor. 
 - Go to "users page"
 - Select "Fantasiløs Spennende Tiger" -> "Standhafig Oldemor" 
 - Select the role tab 
Skjermbilde 2025-11-19 kl  14 46 33

Via Parent: 
 - log in with (20838198385) and choose "Diskret Nær Tiger AS" as the actor. 
 - Go to "reportee page"
 - Select "Blaut Økonomisk Isbjørn SA" -> "Blaut Økonomisk Isbjørn SA **sub-unit**" 
 - Select the role tab 
Skjermbilde 2025-11-19 kl  14 47 47

Via agent: 
Not sure how to test this, but it should be visible in storybook as a test case. 
http://localhost:6006/?path=/story/features-amui-delegationmodal-roleinfo--via-agent

## Summary by CodeRabbit

* **New Features**
  * Structured access-status messages for English, Bokmål and Nynorsk clarifying access via role, parent org, or agent.
  * Mocked role API endpoints for local development/testing.
  * New Storybook scenarios to preview role information and status variations.

* **Refactor**
  * Simplified role info UI with a consolidated RoleStatusMessage component and streamlined status-derivation logic.

### DIFF
--- a/src/features/amUI/common/DelegationModal/Role/RoleStatusMessages.tsx
+++ b/src/features/amUI/common/DelegationModal/Role/RoleStatusMessages.tsx
@@ -42,9 +42,9 @@ export const RoleStatusMessage = ({ role }: RoleStatusMessageProps) => {
 
   const direction = actingParty?.partyUuid === fromParty?.partyUuid ? 'from' : 'to';
   const formattedUserName = formatDisplayName({
-    fullName: (direction === 'from' ? fromParty?.name : toParty?.name) || '',
+    fullName: (direction === 'from' ? toParty?.name : fromParty?.name) || '',
     type:
-      (direction === 'from' ? fromParty?.partyTypeName : toParty?.partyTypeName) ===
+      (direction === 'from' ? toParty?.partyTypeName : fromParty?.partyTypeName) ===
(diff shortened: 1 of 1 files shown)


allinox (approved):
Thank you so much for the detailed description in the PR! 🙌
It helped a ton with understanding the cases and reasoning of the code.

Works like a charm too! 😄

PR #1822, opened 2026-01-08
Allow for choice of multiple service owners in the resourceFilterToolbar

allinox:
image

## Description
This change was mainly motivated by me finding it annoying to have to click on the filter for serviceowner twice (once to activate the filter and then again to choose which filter to use even though we only have the one filter) 😅. The only way I found to fix this is to use multiselect for the filter options and then make the filter non-removable. This allows the user to choose multiple service owners at the same time, but then, why not allow them to do that?

In the long term, I hope we can improve the filter component in the altinn-components library to allow for the case where a single filter can be removable on its own without having this extra level of "which filter would you like to use" when there is only one.

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Owner filter supports selecting multiple owners and shows all selected owners in the label.
  * Owner filter is now fixed in the toolbar and cannot be removed.

* **Bug Fixes**
  * Filtering now reliably handles missing/empty owner selection without errors.

* **Style**
  * Resource list container spacing and max-height behavior improved for more consistent layout.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/features/amUI/common/ResourceList/ResourceList.module.css
+++ b/src/features/amUI/common/ResourceList/ResourceList.module.css
@@ -16,7 +16,7 @@
 }
 
 .resourceListContainer {
-  padding-top: 4px;
+  padding-top: var(--ds-size-1);
 }
 
 .resourceListContainer.maxHeight {


src/features/amUI/common/ResourceList/ResourceList.module.css
```diff
@@ -15,7 +15,12 @@
   width: 100%;
 }
 
-.resourceListContainerMaxHeight {
+.resourceListContainer {
+  padding-top: 4px;
```
mgunnerud: Nitpick: can we use `--ds-size-x` variables for new small padding/margins?

PR #2112, opened 2026-04-13
Backend: add dialog lookup handling for instance delegations

sonwit:
Isolates the BFF part of the dailog-porten integration update. 

- Adds DialogLookupStatus to DialogLookup : `Success`, `NotFound`, and `Forbidden`.
- Updates InstanceService to map Dialogporten 404 and 403 responses into those statuses instead of dropping the lookup.

https://github.com/Altinn/altinn-authorization-tmp/issues/2676


allinox (approved):
Very promising! 😄
Looking forward to seeing this in action with the GUI!

PR #2149, opened 2026-04-21
Approve/Reject requested packages

allinox:
image

## Description
This PR adds functionality to approve and reject requests of type package, in the same view as you can handle resources.

Packages that the approver does not have themselves cannot be approved, only rejected. (Except if you are a main admin, then you can always approve)

Note that this PR also adds a new hook `useAutoFocusRef` for easy setting of focus to specific elements, such as when changing the content of a modal (like in this case here)

## Related Issue(s)
- https://github.com/Altinn/altinn-authorization-tmp/issues/2858

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Review modal now supports access package requests alongside resource requests; package detail view added.

* **Improvements**
  * Package info reorganized with dedicated header and metadata components.
  * Keyboard focus improvements (auto-focus for back/navigation and modal heading).
  * Resource list visuals tightened (reduced min-height) and new list props for accessibility.

* **Localization**
  * Pluralized strings and added titles/messages for package/resource lists and approval restrictions.

* **Chores**
  * Component library dependency bumped.


src/features/amUI/requestPage/RequestReviewModal/RequestPackageDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+        toPartyName={toPartyName}
+      />
+      {cannotApprove && (
+        <DsAlert
+          data-color='warning'
+          data-size={'xs'}
+        >
+          {t('request_page.cannot_approve_package')}
+        </DsAlert>
+      )}
```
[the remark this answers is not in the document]
allinox: In this case, the cannotApprove condition will already be calculated on mount, so the alert will be present from the beginning and will not change while in this view.

src/features/amUI/requestPage/RequestReviewModal/useRequestReview.ts
```diff
@@ -88,33 +112,66 @@ export const useRequestReview = (request: Request | null, onClose: () => void) =
...
 
   const cannotApprove = useCallback(
-    (resourceId: string) => {
-      const checks = delegationChecks[resourceId];
-      return checks?.some((check) => check.result === false) ?? false;
+    ({ resourceId, packageId }: { resourceId?: string; packageId?: string }) => {
+      if (packageId) {
+        const packageCheck = canDelegatePackage(packageId);
+        console.log('Package delegation check for', packageId, ':', packageCheck);
+        if (packageCheck === undefined) return false; // allow if no data
```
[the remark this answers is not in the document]
allinox: If this package truly cannot be delegated, then backend will stop it. For the UI, we want the user to have a smooth experience, so we display the package as delegable until proven otherwise.

src/features/amUI/requestPage/RequestReviewModal/useRequestReview.ts
```diff
@@ -63,19 +77,29 @@ export const useRequestReview = (request: Request | null, onClose: () => void) =
...
-      !isFetchingRequests &&
+      !isFetchingResourceRequests &&
+      !isFetchingPackageRequests &&
       resourceRequests &&
-      resourceRequests.length > 0 &&
-      snapshotRequests.length === 0
+      packageRequests &&
+      snapshotRequests.resourceRequests.length === 0 &&
+      snapshotRequests.packageRequests.length === 0 &&
+      (resourceRequests.length > 0 || packageRequests.length > 0)
```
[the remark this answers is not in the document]
allinox: Error-handling will be performed in a later PR

src/features/amUI/common/ResourceList/ResourceList.tsx
```diff
@@ -136,18 +140,18 @@ export const ResourceList = <
 
   return (
     <div className={classes.container}>
-      <div className={classes.searchAndAdd}>
-        {enableSearch && resources.length > 0 && (
+      {enableSearch && resources.length > 0 && (
```
sonwit: Image
This code seems to remove the delegate/request button for users that don’t already have access in the UserPage -> SinglerightsSection. Maybe this should be controllable by a prop on the component?

src/features/amUI/requestPage/RequestReviewModal/RequestResourceDetail.tsx
```diff
@@ -29,10 +30,12 @@ export const RequestResourceDetail = ({
   toPartyName,
 }: RequestResourceDetailProps) => {
   const { t } = useTranslation();
+  const backButtonRef = useAutoFocusRef<HTMLButtonElement>();
```
sonwit, line 33: Should we be consistent with placing the focus on the header within the modal? 
I honestly don’t know what the best approach here is, but placing the focus on the heading might give a better indication to the user what the content is showing than placing it on the back button? 😅 

i found this on [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
> If the dialog content includes semantic structures, such as lists, tables, or multiple paragraphs, that need to be perceived in order to easily understand the content, i.e., if the content would be difficult to understand when announced as a single unbroken string, then it is advisable to add tabindex="-1" to a static element at the start of the content and initially focus that element. This makes it easier for assistive technology users to read the content by navigating the semantic structures. Additionally, it is advisable to omit applying aria-describedby to the dialog container in this scenario.

src/features/amUI/requestPage/RequestReviewModal/RequestPackageDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+  processedStatus,
+  actionLoading,
+  onBack,
+  onApprove,
+  onReject,
+  cannotApprove,
+  toPartyName,
+}: RequestPackageDetailProps) => {
+  const { t } = useTranslation();
+  const backButtonRef = useAutoFocusRef<HTMLButtonElement>();
```
sonwit, line 34: https://github.com/Altinn/altinn-access-management-frontend/pull/2149/changes#r3118031016

sonwit (approved):
Nice work on fixing this. Works like a charm. 
I only have a few comments, but overall it’s looking good 👍🏼 
✅

src/features/amUI/common/ResourceList/ResourceList.tsx
allinox: Well spotted!

And you're right, of course. I'll fix! 🔨

src/features/amUI/requestPage/RequestReviewModal/RequestResourceDetail.tsx
allinox: I think this is really tricky honestly.

The reason why I chose to place focus on the back button is because I was worried that if we skipped it (by going directly to the heading) then the user wouldn't know the back button was there.

This article on W3C also says

> It is strongly recommended that the tab sequence of all dialogs include a visible element with role button that closes the dialog, such as a close icon or cancel button.

This might be why the Design system by default places focus on the close button when opening their modals. I think that in this case (where you are on a second view inside a single modal) the back button functions as such a "close" button for this second view. And if we place the focus on the title, the back button will no longer appear in direct sequence from the starting point.

So I don't disagree with the wish for consistency or context the header provides as a starting point, but I am worried it will cause the user to miss something important.

If anything.. The fact that we set focus to the title after returning to the "primary view" of the request modal, instead of to the element which "opened" the secondary view... maybe _that_ might be the bigger issue. 😅

allinox: We'll fix this in another PR I think. Since this is a pattern we need to figure out for many parts of our solution

PR #2188, opened 2026-05-05
Set error-snackbars to default duration

allinox:
## Description

After a request from UX (and I agree) we change the error-toasts to no longer be infinite duration. They do not give a lot of information anyway so the value of leaving them there to be screenshotted is minimal. Also, since they do appear to disappear when closing the modal that owns them only to then appear again if you open that modal again, they currently make the solution feel buggy.

For consistency, we use the same (default) duration on ALL snackbars

## Related Issue(s)
- https://github.com/Altinn/altinn-authorization-tmp/issues/2932

## Verification
- [ ] All tests run green

## Documentation

## Release Notes

* **Bug Fixes**
  * Error notification messages now automatically dismiss after a default time period instead of remaining on screen indefinitely across access package and request management features.


sonwit (approved):
Looks good 👍🏼

PR #2323, opened 2026-06-19
Removed depricated A2Link code and reference to old solution

allinox:
## Description

Now that the hideA2Links feature flag is active in all environments, the flag and the code it hides, is no longer needed.
This PR removes this deprecated/unused code.
Also, it removes a phrase that was present on A2 and ER roles that informed the user that the role might give access to something in the old solution. That is no longer accurate.

image
How it looked before this change 👆
How it looks now 👇 
image

Otherwise, everything should be the same

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **Chores**
  * Removed legacy Altinn 2 UI elements, including old roles/consent alerts and the legacy resources disclaimer and links.
  * Removed the “leave beta” shortcut from navigation.
  * Simplified Party Representation error/alert handling and removed the unsynced/invalid connection alert flow.
  * Updated feature-flag configuration by removing the old flag and relying on the new one across environment settings.
* **Tests**
  * Updated Party Representation context tests to match the revised error-handling and link rendering behavior.


sonwit (approved):
Ser bra ut 👍🏼

PR #2327, opened 2026-06-22
Display migrated correspondances if they are not marked as deprecated

allinox:
image

## Description

Before, we had three cases that marked a resource as "Expired":
1. The resource is a migrated app
2. The resource is a migrated correspondance
3. The resource has status "Deprecated" in Resource Registry

Now, all migrated correspondances have automatically been set to status "Deprecated" and service owners are encouraged to change this if they do not agree. Thus, we want to remove case nr 2 so that service owners are free to choose to mark even old correspondace services as active if they want.

## Related Issue(s)
- https://github.com/Altinn/altinn-authorization-tmp/issues/3604

## Verification
- [ ] All tests run green

## Documentation

## Release Notes

* **Bug Fixes**
  * Fixed resource expiration logic: Resources are now correctly identified as expired based on their status field (when "deprecated") or resource type, rather than identifier patterns. This change ensures active migrated resources appear correctly in search results and enables more accurate filtering when excluding expired resources from the application.


sonwit (approved):
Looks good 👍🏼

PR #2348, opened 2026-06-29
GetReportee returns the actual subunit when reportee is a subunit

allinox:
## Description

The new enduser endpoint does not work quite the same as the old reportee endpoint did, and it always returns the full parent party even if the filter requested a subunit. Thus, we have to double-check if we actually got the party we requested and fetch the party from the subunits if not.

This also fixes the currently failing playwright test

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Party lookups now validate returned results and only return the exact requested party, including subunit matching when the party UUID doesn’t match.
  * Improved behavior prevents unrelated or incorrect entries from being returned.
  * Added automated test coverage for subunit-based party retrieval to ensure consistent results.
* **Chores**
  * Updated `@altinn/altinn-components` to the latest version.


mgunnerud (approved):
🙌

